PR #230, opened 2022-03-22
Shuffle sample list

### LINKED ISSUE #194: Randomly apply manipulator when applying manipulator with limit
ArbitraryNodes found by expression has fixed ordering.
So applying manipulator with parameter `limit`, which parameter used as limiting the number of applied ArbitraryNode, always returns same applied ArbitraryNode.
It is not an intuitive result.

In short, this issue is to shuffle all `ArbitraryNodes` found by `ArbitraryExpression`.

-------

적용 회수를 입력받는 연산 (limit 을 입력받는 연산)에서 조회한 노드 순서대로 적용하는 게 아니라 랜덤하게 적용하도록 수정합니다.

sandrawangyx:
This PR is to fix #194

### DIFF
--- a/fixture-monkey/src/main/java/com/navercorp/fixturemonkey/arbitrary/ArbitraryTree.java
+++ b/fixture-monkey/src/main/java/com/navercorp/fixturemonkey/arbitrary/ArbitraryTree.java
@@ -22,6 +22,7 @@ import static com.navercorp.fixturemonkey.Constants.HEAD_NAME;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -63,6 +64,9 @@ public final class ArbitraryTree<T> {
 			selectNodes.addAll(nextNodes);
 			nextNodes.clear();
 		}
+		List selectNodesList = new ArrayList(selectNodes);
+		Collections.shuffle(selectNodesList);
+		selectNodes = new LinkedList<>(selectNodesList);
 		return selectNodes;
 	}
 


sandrawangyx:
hi @seongahjo,  plz take a look at this PR.
Btw, is there any formatter enforcement such as `spotless` as well as tools to detect code smell that is being used?

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/arbitrary/ArbitraryTree.java
```diff
@@ -63,6 +64,9 @@ public ArbitraryTree(ArbitraryNode<T> head) {
 			selectNodes.addAll(nextNodes);
 			nextNodes.clear();
 		}
+		List selectNodesList = new ArrayList(selectNodes);
+		Collections.shuffle(selectNodesList);
+		selectNodes = new LinkedList<>(selectNodesList);
```
seongahjo, line 69: I think it is better to switch `selectNodes` as `LinkedList` from `Queue`.
Because `selectNodes` should be an implementation of `Queue` and `List` in this case.

It is better to use `Random` instance looked up by `Randoms` for reproducing same results when seed is fixed

seongahjo:
@sandrawangyx 
We`re currently using `Checkstyle` instead of `spotless`

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/arbitrary/ArbitraryTree.java
```diff
@@ -22,20 +22,23 @@
...
 import javax.annotation.Nullable;
 
 import net.jqwik.api.Arbitrary;
 
+import com.navercorp.fixturemonkey.api.random.Randoms;
 import com.navercorp.fixturemonkey.generator.ArbitraryGenerator;
 import com.navercorp.fixturemonkey.validator.ArbitraryValidator;
 
+
+
```
seongahjo, line 41: Blank lines should be removed if it is not necessary

mhyeon-lee (approved):

seongahjo:
Any idea for testing if ArbitraryNodes are randomly shuffled?

acktsap (approved):

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/arbitrary/ArbitraryTree.java
sandrawangyx: I could address this in a separate PR.

sandrawangyx:
it is probably hard to test the true randomness of the nodes in the ArbitraryTree. I was thinking that we could test that if Collections.shuffle() is being called, what we can do is to reply on the Mockito to mock the invocations to the shuffle method calls. Let me know your thoughts.
I can address this in a different PR.
BTW, if the PR is approved, should I be the one who merges it or someone else from the reviewers' group should merge it?

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/arbitrary/ArbitraryTree.java
seongahjo: @sandrawangyx 
Are there any reasons why you remove blank lines in separate PR?
Since we use `Squash and merge`, you do not need to care about commit logs.

seongahjo:
@sandrawangyx 
Yes, I agree.
It is hard to test if ArbitraryNodes are shuffled or not.
I think using `mockito` for testing makes a fragile test.
Because it is likely to change the way of randomizing ArbitraryNode instead of using `Collections.shuffle()` in near future.
So I think better not test than test calling `Collections.shuffle()`

| someone else from the reviewers' group should merge it
We'll merge PR as soon as possible.

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/arbitrary/ArbitraryTree.java
sandrawangyx: cool ty, fix now

sandrawangyx:
@seongahjo 
I found a couple of help wanted tagged issues in the list. 
I could help out with the next one if there is something that needs to be looked at immediately.

Issue #487, opened 2022-11-12
EnumMap isn't generated if the number of enum values is less than container min size


esfomeado:
If the container size is bigger then the number of enum values it will fail to generate the `EnumMap`.

This is a special case that should ignore the min container size.

Sample project:
https://github.com/esfomeado/monkey-enum-bug

Two tests are provided:

` void testWillAlwaysFail()` - It will fail always since the container size is bigger than the number of enum values.

`void testSometimesFail()` - It will only fail sometimes since sometimes the container size is the same as the number of enum values.

seongahjo:
@esfomeado 
Thank you for reporting bug!

It would be fixed since 0.4.4
https://github.com/naver/fixture-monkey/pull/488

esfomeado:
@seongahjo I've tested and it's working.

Thanks for the fix.

PR #488, opened 2022-11-12
Fix generating enum set or map

seongahjo:
## Features
EnumMap or EnumSet size which generated by FixtureMonkey  would not exceed enum size if no `size` manipulation

## Issue
https://github.com/naver/fixture-monkey/issues/487


mhyeon-lee (approved):

mhyeon-lee:
bug-fix 로 0.4.4 바로 가도 괜찮을거 같습니다.

PR #555, opened 2023-01-28
Add a condition to not include control characters when creating String values (#553)

### LINKED ISSUE #553: Test code intermittently fails when using JavaxValidationPlugin
## Describe the bug
Test code intermittently fails when using `JavaxValidationPlugin`

## Your environment
* version of Fixture Monkey: 0.4.11
* version of Java/Kotlin: Kotlin 1.7.22

## Steps to reproduce
[Repository](https://github.com/doljae/monkey-fixture) that reproduced the problem

I wrote [2 test codes](https://github.com/doljae/monkey-fixture/blob/main/src/test/kotlin/com/example/monkeyfixture/TestControllerTest.kt). This test code does the same logic. Only the presence or absence of `LabMonkey's` `JavaxValidationPlugin` that creates fixtures is different.

```kotlin
@Test
fun getSavedQuestionnaire() {
    val uri = "/test"
    val sut: LabMonkey = LabMonkey.labMonkeyBuilder()
        .defaultNullInjectGenerator { DefaultNullInjectGenerator.NOT_NULL_INJECT }
        .plugin(KotlinPlugin())
        .plugin(JacksonPlugin())
        .build()
    val fixture = sut.giveMeOne<TestDataClass>()
    println(fixture.toString())
    every { testService.test() } returns fixture

    mockMvc
        .perform(MockMvcRequestBuilders.get(uri))
        .andExpect(MockMvcResultMatchers.status().isOk)
        .andExpect(MockMvcResultMatchers.jsonPath("$.id").value(fixture.id))
        .andExpect(MockMvcResultMatchers.jsonPath("$.name").value(fixture.name))
        .andExpect(MockMvcResultMatchers.jsonPath("$.name2").value(fixture.name2))
        .andExpect(MockMvcResultMatchers.jsonPath("$.name3").value(fixture.name3))
        .andDo(MockMvcResultHandlers.print())
}

// Fails intermittently
@Test
fun getSavedQuestionnaireWithJavaxPlugin() {
    val uri = "/test"
    val sut: LabMonkey = LabMonkey.labMonkeyBuilder()
        .defaultNullInjectGenerator { DefaultNullInjectGenerator.NOT_NULL_INJECT }
        .plugin(KotlinPlugin())
        .plugin(JacksonPlugin())
        .plugin(JavaxValidationPlugin())
        .build()
    val fixture = sut.giveMeOne<TestDataClass>()
    println(fixture.toString())
    every { testService.test() } returns fixture

    mockMvc
        .perform(MockMvcRequestBuilders.get(uri))
        .andExpect(MockMvcResultMatchers.status().isOk)
        .andExpect(MockMvcResultMatchers.jsonPath("$.id").value(fixture.id))
        .andExpect(MockMvcResultMatchers.jsonPath("$.name").value(fixture.name))
        .andExpect(MockMvcResultMatchers.jsonPath("$.name2").value(fixture.name2))
        .andExpect(MockMvcResultMatchers.jsonPath("$.name3").value(fixture.name3))
        .andDo(MockMvcResultHandlers.print())
}
```

## Expected behaviour
`JavaxValidation` plug-in was added, but there is no part where `javax.validation`-related logic is used in fixture.
Therefore, I expect the same result, i.e. pass both test cases.

## Actual behaviour
Test cases using `JavaxValidationPlugin` fail intermittently.

## Additional info
It could be a bug in Spring and related tools, or my lack of proficiency in using it
Not exactly accurate, but the output of the path seems to be a Unicode-related issue.

doljae:
## Summary
Fix #553 

## (Optional): Description
Add a condition to not include [control characters](https://en.wikipedia.org/wiki/Control_character) when creating String values through `XXXValidationJavaArbitraryResolver`.

## How Has This Been Tested?
Add additional assertions to existing test code

### DIFF
--- a/fixture-monkey-jakarta-validation/src/main/java/com/navercorp/fixturemonkey/jakarta/validation/introspector/JakartaValidationJavaArbitraryResolver.java
+++ b/fixture-monkey-jakarta-validation/src/main/java/com/navercorp/fixturemonkey/jakarta/validation/introspector/JakartaValidationJavaArbitraryResolver.java
@@ -50,7 +50,6 @@ import com.navercorp.fixturemonkey.api.introspector.JavaArbitraryResolver;
 
 @API(since = "0.4.10", status = Status.EXPERIMENTAL)
 public final class JakartaValidationJavaArbitraryResolver implements JavaArbitraryResolver {
-	private static final String CONTROL_BLOCK = "\u0000-\u001f\u007f";
 	private static final RegexGenerator REGEX_GENERATOR = new RegexGenerator();
 
 	private final JakartaValidationConstraintGenerator constraintGenerator;
@@ -117,15 +116,29 @@ public final class JakartaValidationJavaArbitraryResolver implements JavaArbitra
 		return arbitrary
 			.filter(it -> {
 				if (!notBlank) {
-					return true;
+					if (it == null) {
+						return true;
+					} else {
+						return !containsControlCharacters(it);
+					}
 				}
 
-				if (it == null || it.trim().length() == 0) {
+				if (it == null || it.trim().isEmpty()) {
 					return false;
 				}
 
-				return !CONTROL_BLOCK.equals(it.trim());
+				return !containsControlCharacters(it);
 			});
+	}
+
+	private static boolean containsControlCharacters(String value) {
+		for (char c : value.toCharArray()) {
+			if (Character.isISOControl(c)) {
+				return true;
+			}
+		}
+
+		return false;
 	}
 
 	@Override
--- a/fixture-monkey-jakarta-validation/src/test/java/com/navercorp/fixturemonkey/jakarta/validation/JakartaValidationFixtureMonkeyTest.java
+++ b/fixture-monkey-jakarta-validation/src/test/java/com/navercorp/fixturemonkey/jakarta/validation/JakartaValidationFixtureMonkeyTest.java
@@ -227,10 +227,6 @@ class JakartaValidationFixtureMonkeyTest {
 		then(actual.getEmail()).containsOnlyOnce("@");
 		Pattern controlCharacters = Pattern.compile("[\u0000-\u001f\u007f]");
 		then(actual.getStr()).doesNotMatch(controlCharacters);
-		then(actual.getNotEmpty()).doesNotMatch(controlCharacters);
-		then(actual.getSize()).doesNotMatch(controlCharacters);
-		then(actual.getDigits()).doesNotMatch(controlCharacters);
-		then(actual.getEmail()).doesNotMatch(controlCharacters);
 	}
 
 	@Property(tries = 100)
--- a/fixture-monkey-jakarta-validation/src/test/java/com/navercorp/fixturemonkey/jakarta/validation/JakartaValidationFixtureMonkeyTest.java
+++ b/fixture-monkey-jakarta-validation/src/test/java/com/navercorp/fixturemonkey/jakarta/validation/JakartaValidationFixtureMonkeyTest.java
@@ -225,6 +225,12 @@ class JakartaValidationFixtureMonkeyTest {
 			then(ch).isBetween('e', 'o');
 		}
 		then(actual.getEmail()).containsOnlyOnce("@");
+		Pattern controlCharacters = Pattern.compile("[\u0000-\u001f\u007f]");
+		then(actual.getStr()).doesNotMatch(controlCharacters);
+		then(actual.getNotEmpty()).doesNotMatch(controlCharacters);
+		then(actual.getSize()).doesNotMatch(controlCharacters);
+		then(actual.getDigits()).doesNotMatch(controlCharacters);
+		then(actual.getEmail()).doesNotMatch(controlCharacters);
 	}
 
 	@Property(tries = 100)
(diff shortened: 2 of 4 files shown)


fixture-monkey-jakarta-validation/src/main/java/com/navercorp/fixturemonkey/jakarta/validation/introspector/JakartaValidationJavaArbitraryResolver.java
```diff
@@ -117,17 +116,31 @@ public Arbitrary<String> strings(
...
+	private static boolean containsControlCharacters(String value) {
+		for (char c : value.toCharArray()) {
+			if (Character.isISOControl(c)) {
+				return true;
+			}
+		}
+
+		return false;
+	}
+
```
doljae, line 143: This is a private method that uses a built-in function of the `Character` class to return whether this character is a control character.

fixture-monkey-jakarta-validation/src/main/java/com/navercorp/fixturemonkey/jakarta/validation/introspector/JakartaValidationJavaArbitraryResolver.java
```diff
@@ -117,17 +116,31 @@ public Arbitrary<String> strings(
 		return arbitrary
 			.filter(it -> {
 				if (!notBlank) {
-					return true;
+					if (it == null) {
+						return true;
+					} else {
+						return !containsControlCharacters(it);
+					}
```
doljae, line 123: I implemented it in the simplest way possible, but if there is a better way or any recommendation, feel free to suggest it 🙂

fixture-monkey-jakarta-validation/src/test/java/com/navercorp/fixturemonkey/jakarta/validation/JakartaValidationFixtureMonkeyTest.java
```diff
@@ -225,6 +225,12 @@ void sampleString() {
 			then(ch).isBetween('e', 'o');
 		}
 		then(actual.getEmail()).containsOnlyOnce("@");
+		Pattern controlCharacters = Pattern.compile("[\u0000-\u001f\u007f]");
+		then(actual.getStr()).doesNotMatch(controlCharacters);
+		then(actual.getNotEmpty()).doesNotMatch(controlCharacters);
+		then(actual.getSize()).doesNotMatch(controlCharacters);
+		then(actual.getDigits()).doesNotMatch(controlCharacters);
+		then(actual.getEmail()).doesNotMatch(controlCharacters);
```
SooKim1110, line 233: Asserting `str` seems to be enough. Other fields such as `notEmpty` are there to check if a valid object is made given the specific annotation.

doljae: Thanks for the review 🙂 
However, the String field with `NotEmpty` annotation has also been modified so that control characters are not entered according to the current changes. So, the test code above seems to be meaningful. What do you think?

Please let me know if there is something I am misunderstanding.

doljae: @SooKim1110 ,
And when I re-requested the review, another maintainer disappeared from the reviewer 😭

SooKim1110: Do you mean the `NotBlank` annotation? Then I personally think that it's okay if you just check `str` and `notBlank` because the code you changed in `JakartaValidationJavaArbitraryResolver` filters differently according to the presence of that annotation. Any other thoughts @seongahjo ?

SooKim1110: No worries! I'll put them back

doljae: Sorry I was going to talk about `NotBlank`.
In commit 6a77ec0, I only added assertions about `str` and `notBlank`.

seongahjo: @doljae 
I think there is no need to check `@NotBlank` if generates a control block character.
This pr affects String generated by `@NotBlank`, this is because it affects all String generated by StringArbitrary.
`@NotBlank` only checks if there is a non-whitespace character.

In my personal opinion, a simplest assertion is needed in order to tell the reader how it may change directly, preventing a false alarm.

cc @SooKim1110

doljae: @seongahjo 
I agree with your opinion.
In commit ea916ac, I left assertions for `str` only 🙂

seongahjo (approved):

SooKim1110 (approved):

seongahjo:
@doljae 
Thank you for your contributing!

PR #581, opened 2023-02-18
Replace defaultArbitraryContainerInfo to defaultArbitraryContainerInfoGenerator

### LINKED ISSUE #569: When adding the jakarta-validation plugin, the defaultArbitraryContainerInfo setting does not work properly
## Describe the bug
When adding the `jakarta-validation` plugin, the `defaultArbitraryContainerInfo` setting does not work properly

## Your environment
* version of Fixture Monkey: 0.4.13
* version of Java/Kotlin: Kotlin 1.7.22

## Steps to reproduce
[Repository](https://github.com/doljae/monkey-fixture) that reproduced the problem

I wrote [2 test codes](https://github.com/doljae/monkey-fixture/blob/main/src/test/kotlin/com/example/monkeyfixture/ContainerTest.kt). This test code does the same logic. Only the presence or absence of `LabMonkey's` `JakartaValidationPlugin` that creates fixtures is different.

```kotlin
// ==== data class ====
data class TestDataClassV2(
    val id: Long,
    val name: String,
    val container: List<String>,
    val container2: List<InnerTestDataClass>
) {
    data class InnerTestDataClass(val id: Long, val name: String, val name2: String, val name3: String)
}

// ==== test cases ====
@RepeatedTest(1000)
fun test() {
    val minSize = 3
    val maxSize = 10
    val sut: LabMonkey = LabMonkey.labMonkeyBuilder()
        .plugin(KotlinPlugin())
        .plugin(
            JacksonPlugin(
                jacksonObjectMapper()
                    .registerModule(JavaTimeModule())
                    .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
                    .setDefaultPropertyInclusion(JsonInclude.Include.NON_NULL)
            )
        )
        .defaultArbitraryContainerInfo(ArbitraryContainerInfo(minSize, maxSize, false))
        .defaultNullInjectGenerator { DefaultNullInjectGenerator.NOT_NULL_INJECT }
        .build()

    val giveMeOne = sut.giveMeOne<TestDataClassV2>()
    Assertions.assertThat(giveMeOne.container.size >= minSize).isTrue
    Assertions.assertThat(giveMeOne.container.size <= maxSize).isTrue
    Assertions.assertThat(giveMeOne.container2.size >= minSize).isTrue
    Assertions.assertThat(giveMeOne.container2.size <= maxSize).isTrue
}

// fail
@RepeatedTest(1000)
fun testWithJakartaValidationPlugin() {
    val minSize = 3
    val maxSize = 10
    val sut: LabMonkey = LabMonkey.labMonkeyBuilder()
        .plugin(KotlinPlugin())
        .plugin(
            JacksonPlugin(
                jacksonObjectMapper()
                    .registerModule(JavaTimeModule())
                    .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
                    .setDefaultPropertyInclusion(JsonInclude.Include.NON_NULL)
            )
        )
        .plugin(JakartaValidationPlugin())
        .defaultArbitraryContainerInfo(ArbitraryContainerInfo(minSize, maxSize, false))
        .defaultNullInjectGenerator { DefaultNullInjectGenerator.NOT_NULL_INJECT }
        .build()

    val giveMeOne = sut.giveMeOne<TestDataClassV2>()
    Assertions.assertThat(giveMeOne.container.size >= minSize).isTrue
    Assertions.assertThat(giveMeOne.container.size <= maxSize).isTrue
    Assertions.assertThat(giveMeOne.container2.size >= minSize).isTrue
    Assertions.assertThat(giveMeOne.container2.size <= maxSize).isTrue
}
```

## Expected behaviour
I expect the same result, i.e. pass both test cases.

## Actual behaviour
Test case using `JakartaValidationPlugin` fail.

## Additional info
If this is intentional, please comment 🙂

doljae:
## Summary
Fix #569 

## (Optional): Description
Replace `defaultArbitraryContainerInfo` to `defaultArbitraryContainerInfoGenerator`

## How Has This Been Tested?
None

## Additional info
The maintainers [provided guidelines](https://github.com/naver/fixture-monkey/issues/569#issuecomment-1428916871) for doing this.
Unfortunately, I don't think I understood this properly. I worked on it first, but I'm honestly not sure if my work went in the right direction 😞 

Regarding this bug, I don't think it's good for a discovered bug to remain for a long time.
So it would be nice to proceed with fixing the bug at the maintainer's discretion.

### DIFF
--- a/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/option/GenerateOptionsBuilder.java
+++ b/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/option/GenerateOptionsBuilder.java
@@ -34,7 +34,6 @@ import org.apiguardian.api.API;
 import org.apiguardian.api.API.Status;
 
 import com.navercorp.fixturemonkey.api.customizer.FixtureCustomizer;
-import com.navercorp.fixturemonkey.api.generator.ArbitraryContainerInfo;
 import com.navercorp.fixturemonkey.api.generator.ArbitraryContainerInfoGenerator;
 import com.navercorp.fixturemonkey.api.generator.ArbitraryGenerator;
 import com.navercorp.fixturemonkey.api.generator.ContainerPropertyGenerator;
@@ -71,8 +70,7 @@ public final class GenerateOptionsBuilder {
 	private List<MatcherOperator<NullInjectGenerator>> nullInjectGenerators = new ArrayList<>();
 	private NullInjectGenerator defaultNullInjectGenerator;
 	private List<MatcherOperator<ArbitraryContainerInfoGenerator>> arbitraryContainerInfoGenerators = new ArrayList<>();
-	private Integer defaultArbitraryContainerMaxSize;
-	private ArbitraryContainerInfo defaultArbitraryContainerInfo;
+	private ArbitraryContainerInfoGenerator defaultArbitraryContainerInfoGenerator;
 	private List<MatcherOperator<ArbitraryGenerator>> arbitraryGenerators = new ArrayList<>();
 	private ArbitraryGenerator defaultArbitraryGenerator;
 
@@ -311,13 +309,9 @@ public final class GenerateOptionsBuilder {
 		);
 	}
 
-	public GenerateOptionsBuilder defaultArbitraryContainerMaxSize(int defaultArbitraryContainerMaxSize) {
-		this.defaultArbitraryContainerMaxSize = defaultArbitraryContainerMaxSize;
-		return this;
-	}
-
-	public GenerateOptionsBuilder defaultArbitraryContainerInfo(ArbitraryContainerInfo defaultArbitraryContainerInfo) {
-		this.defaultArbitraryContainerInfo = defaultArbitraryContainerInfo;
+	public GenerateOptionsBuilder defaultArbitraryContainerInfoGenerator(
+		ArbitraryContainerInfoGenerator defaultArbitraryContainerInfoGenerator) {
+		this.defaultArbitraryContainerInfoGenerator = defaultArbitraryContainerInfoGenerator;
 		return this;
 	}
 
@@ -529,15 +523,8 @@ public final class GenerateOptionsBuilder {
 			defaultNullInjectGenerator = defaultNullInjectGeneratorOperator.apply(defaultNullInjectGenerator);
 		}
 
-		int defaultArbitraryContainerMaxSize = defaultIfNull(
-			this.defaultArbitraryContainerMaxSize,
-			() -> GenerateOptions.DEFAULT_ARBITRARY_CONTAINER_MAX_SIZE
-		);
-		ArbitraryContainerInfo defaultArbitraryContainerInfo =
-			defaultIfNull(
-				this.defaultArbitraryContainerInfo,
-				() -> new ArbitraryContainerInfo(0, defaultArbitraryContainerMaxSize)
-			);
+		ArbitraryContainerInfoGenerator defaultArbitraryContainerInfoGenerator =
+			this.defaultArbitraryContainerInfoGenerator;
 		ArbitraryGenerator defaultArbitraryGenerator =
 			defaultIfNull(this.defaultArbitraryGenerator, this.javaDefaultArbitraryGeneratorBuilder::build);
 
@@ -552,8 +539,7 @@ public final class GenerateOptionsBuilder {
 			this.nullInjectGenerators,
 			defaultNullInjectGenerator,
 			this.arbitraryContainerInfoGenerators,
-			defaultArbitraryContainerMaxSize,
-			defaultArbitraryContainerInfo,
+			defaultArbitraryContainerInfoGenerator,
 			this.arbitraryGenerators,
 			defaultArbitraryGenerator,
 			this.arbitraryCustomizers,
--- a/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/option/GenerateOptionsBuilder.java
+++ b/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/option/GenerateOptionsBuilder.java
@@ -34,6 +34,7 @@ import org.apiguardian.api.API;
 import org.apiguardian.api.API.Status;
 
 import com.navercorp.fixturemonkey.api.customizer.FixtureCustomizer;
+import com.navercorp.fixturemonkey.api.generator.ArbitraryContainerInfo;
 import com.navercorp.fixturemonkey.api.generator.ArbitraryContainerInfoGenerator;
 import com.navercorp.fixturemonkey.api.generator.ArbitraryGenerator;
 import com.navercorp.fixturemonkey.api.generator.ContainerPropertyGenerator;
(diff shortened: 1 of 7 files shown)


fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/option/GenerateOptionsBuilder.java
```diff
@@ -529,15 +523,8 @@ public GenerateOptions build() {
...
-			this.defaultArbitraryContainerMaxSize,
-			() -> GenerateOptions.DEFAULT_ARBITRARY_CONTAINER_MAX_SIZE
-		);
-		ArbitraryContainerInfo defaultArbitraryContainerInfo =
-			defaultIfNull(
-				this.defaultArbitraryContainerInfo,
-				() -> new ArbitraryContainerInfo(0, defaultArbitraryContainerMaxSize)
-			);
+		ArbitraryContainerInfoGenerator defaultArbitraryContainerInfoGenerator =
+			this.defaultArbitraryContainerInfoGenerator;
```
seongahjo, line 527: If `defaultArbitraryContainerInfo` is not set (null), you should set `defaultArbitraryContainerInfoGenerator`.
ArbitraryContainerInfo in `ContainerPropertyGeneratorContext` means  a result of `size` manipulation.

You should set ArbitraryContainerInfoGenerator returns defaultArbitraryContainerInfo.
check out below.
```
ArbitraryContainerInfo defaultArbitraryContainerInfo =
			defaultIfNull(
				this.defaultArbitraryContainerInfo,
				() -> new ArbitraryContainerInfo(0, defaultArbitraryContainerMaxSize)
			);
```

fixture-monkey/src/test/java/com/navercorp/fixturemonkey/test/InnerSpecTest.java
```diff
@@ -136,7 +136,7 @@ void keyNullThrows() {
 	@Property
 	void keyInKey() {
 		FixtureMonkey sut = FixtureMonkey.builder()
-			.defaultArbitraryContainerInfo(new ArbitraryContainerInfo(1, 3))
+			.defaultArbitraryContainerInfoGenerator(ContainerPropertyGeneratorContext::getContainerInfo)
```
seongahjo, line 139: You should not map getContaienrInfo since it is a result of a `size` manipulation.
If there is no size manipulation, throws NPE

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/option/GenerateOptionsBuilder.java
doljae: I don't understand `ArbitraryContainerInfoGenerator returns defaultArbitraryContainerInfo.`

As you said, I removed `defaultArbitraryContainerInfo` field in `GenerateOptions`. So, we don't have to deal with the `defaultArbitraryInfo` field in `GenerateOptionsBuilder` from my point of view 🤔 

And to get `ArbitraryInfo` returned by `ArbitraryContainerInfoGenerator.generate()`, `ContainerPropertyGeneratorContext` is required, but there seems to be no place to use it inside the `build()` method.

doljae: And to use the code you commented, it seems that the `defaultArbitraryContainerInfo` and `defaultArbitraryContainerMaxSize` fields should be present as fields of `GenerateOptionsBuilder`, but these two fields have also been removed from `GenerateOptions`.

It's clear that I'm not understanding it properly 😭

seongahjo: Yes, there is no `defaultArbitraryContainerInfo` option now.
I mean `defaultArbitraryContainerInfo` as an object of `ArbitraryContainerInfo`.

Interface `ArbitraryContainerInfoGenerator` should return `ArbitraryContainerInfo`.
```java
@FunctionalInterface
public interface ArbitraryContainerInfoGenerator {
	ArbitraryContainerInfo generate(ContainerPropertyGeneratorContext context);
}
```

Since it is functional interface, you don't need `ContainerPropertyGeneratorContext`.
You could use it by lambda expression. 
such as `(ctx) -> new ArbitraryContainerInfo(0, 3)`

`defaultArbitraryContainerInfo` option has a default `ArbitraryContainerInfo` like below.
`defaultArbitraryContainerInfoGenerator` option should have a default `defaultArbitraryContainerInfoGenerator`, same as deprecated `defaultArbitraryContainerInfo` option.

```java
new ArbitraryContainerInfo(0, defaultArbitraryContainerMaxSize)
```

So you have to deal with the situation user does not use `defaultArbitraryContainerInfoGenerator` option.

fixture-monkey/src/test/java/com/navercorp/fixturemonkey/test/InnerSpecTest.java
doljae: Do you have some clues for [this CI result](https://github.com/naver/fixture-monkey/actions/runs/4229897781/jobs/7346691308)?

CI with JDK 11 passed but not JDK 8 🤔

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/option/GenerateOptionsBuilder.java
```diff
@@ -529,15 +524,10 @@ public GenerateOptions build() {
 			defaultNullInjectGenerator = defaultNullInjectGeneratorOperator.apply(defaultNullInjectGenerator);
 		}
 
-		int defaultArbitraryContainerMaxSize = defaultIfNull(
-			this.defaultArbitraryContainerMaxSize,
-			() -> GenerateOptions.DEFAULT_ARBITRARY_CONTAINER_MAX_SIZE
+		ArbitraryContainerInfoGenerator defaultArbitraryContainerInfoGenerator = defaultIfNull(
+			this.defaultArbitraryContainerInfoGenerator,
+			() -> context -> new ArbitraryContainerInfo(0, 3)
```
seongahjo, line 529: `3` could be replaced by `GenerateOptions .DEFAULT_ARBITRARY_CONTAINER_MAX_SIZE`

doljae: In commit e464960, I replaced literal to pre-initialized constant 🙂

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/option/GenerateOptionsBuilder.java
doljae: Thanks for the clear and detailed explanation 👍 
In commit [32dd7be](https://github.com/naver/fixture-monkey/pull/581/commits/32dd7be3838cc34c62052fb10a6c8e182ecd45e4), I tried again according to your guide.

seongahjo (approved):

Issue #713, opened 2023-07-21
JakartaValidationJavaArbitraryResolver doesn't work properly for properties with @Pattern


SooKim1110:
## Describe the bug
The issue occurs when using the JakartaValidationPlugin(or JavaxValidationPlugin) and generating properties with the @Pattern annotation.
There are two problems:

1. `RegexGenerator`
The RegexGenerator class in Fixture monkey replaces certain characters from the given regex, resulting in an invalid regex. For instance, when using @Pattern(regexp = "^[\\\\S]+$"), the "\\\\S" part gets replaced, leading to an invalid regex.

2. `Generex`
Fixture monkey utilizes the Generex library to generate strings that match the given regex. However, Generex can be excessively time-consuming for certain regexes. For example, generating a property with @Pattern(regexp = "\\\\S+") using fixture monkey takes a long time.

## Your environment
* version of Fixture Monkey 
0.6.2
* version of Java/Kotlin 
Java 17

## Steps to reproduce
```java
@Getter
@Setter
public class Sample {
	@Pattern(regexp = "^[\\S]+$")
	private String pattern;
}

private static final FixtureMonkey SUT = FixtureMonkey.builder()
		.plugin(new JakartaValidationPlugin())
		.defaultNotNull(true)
		.build();
```
When using the above class and fixture monkey option, the generation of the class Sample fails with the error message `net.jqwik.api.JqwikException: empty set of values`. This error occurs because no matching string with the given regex is generated, causing the resolver to call Arbitraries.of() with an empty list. 
(An exception handler should be implemented to handle such scenarios)

## Expected behaviour
A string that matches the given pattern should be made

## Actual behaviour
`net.jqwik.api.JqwikException: empty set of values`

PR #870, opened 2024-01-12
Change regular expression generation library to RgxGen

SooKim1110:
## Summary
Related to #772, #713

Changes the regular expression generation library to RgxGen.

## (Optional): Description
- An exception is thrown when RgxGen fails to generate or takes more than 10 seconds.
- The option of generating random values(or null) when RgxGen fails was considered but not implemented because the fixture generation will still fail at the `ArbitaryValidator`
- Some of the cases mentioned in the referenced issues (`^[\\S]+$`, `\\S+`) are now generatable. However, cases like `^(?=.*[a-zA-Z])(?=.*[0-9])(?=.*[!@#$%^&*])[a-zA-Z0-9!@#$%^&*]{8,20}$` fails to be generated because the library has limitations (https://github.com/curious-odd-man/RgxGen?tab=readme-ov-file#limitations). 

## How Has This Been Tested?
Added RegexGeneratorTest.

## Is the Document updated?
Will add a subsequent commit to update the document once this PR is reviewed and finalized.

### DIFF
--- a/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
+++ b/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
@@ -64,10 +64,10 @@ public final class RegexGenerator {
 	}
 
 	public String generate(String regex, int[] flags, Predicate<String> stringCondition) {
-		boolean caseSensitive = Arrays.stream(flags).noneMatch(it -> it == FLAG_CASE_INSENSITIVE);
+		boolean caseInSensitive = Arrays.stream(flags).anyMatch(it -> it == FLAG_CASE_INSENSITIVE);
 
 		try {
-			RgxGen rgxGen = generateRgxGen(regex, caseSensitive);
+			RgxGen rgxGen = generateRgxGen(regex, caseInSensitive);
 
 			String result = executor.submit(() ->
 				rgxGen.stream()
@@ -75,7 +75,7 @@ public final class RegexGenerator {
 					.findFirst()
 			).get(DEFAULT_REGEXP_GENERATION_TIMEOUT_SEC, TimeUnit.SECONDS).orElse(null);
 
-			checkValidity(regex, result, caseSensitive);
+			checkValidity(regex, result, caseInSensitive);
 			return result;
 		} catch (Exception ex) {
 			throw new IllegalArgumentException(
@@ -138,9 +138,9 @@ public final class RegexGenerator {
 		return result;
 	}
 
-	private static RgxGen generateRgxGen(String regex, boolean caseSensitive) {
+	private static RgxGen generateRgxGen(String regex, boolean caseInSensitive) {
 		RgxGenProperties properties = new RgxGenProperties();
-		if (!caseSensitive) {
+		if (caseInSensitive) {
 			RgxGenOption.CASE_INSENSITIVE.setInProperties(properties, true);
 		}
 		RgxGen rgxGen = new RgxGen(regex);
@@ -148,12 +148,12 @@ public final class RegexGenerator {
 		return rgxGen;
 	}
 
-	private static void checkValidity(String regex, String result, boolean caseSensitive) {
+	private static void checkValidity(String regex, String result, boolean caseInSensitive) {
 		Pattern pattern;
-		if (caseSensitive) {
-			pattern = Pattern.compile(regex);
-		} else {
+		if (caseInSensitive) {
 			pattern = Pattern.compile(regex, Pattern.CASE_INSENSITIVE);
+		} else {
+			pattern = Pattern.compile(regex);
 		}
 
 		if (!pattern.matcher(result).matches()) {
--- a/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
+++ b/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
@@ -22,14 +22,13 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
-import java.util.Spliterator;
-import java.util.Spliterators;
+import java.util.NoSuchElementException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
 
 import javax.annotation.Nullable;
 
@@ -39,7 +38,6 @@ import org.apiguardian.api.API.Status;
 import com.github.curiousoddman.rgxgen.RgxGen;
 import com.github.curiousoddman.rgxgen.config.RgxGenOption;
 import com.github.curiousoddman.rgxgen.config.RgxGenProperties;
-import com.github.curiousoddman.rgxgen.iterators.StringIterator;
 
 @API(since = "0.6.9", status = Status.MAINTAINED)
 public final class RegexGenerator {
@@ -48,10 +46,11 @@ public final class RegexGenerator {
 	public static final int DEFAULT_REGEXP_GENERATION_MAX_SIZE = 100;
 
 	public List<String> generateAll(String regex, int[] flags, @Nullable Integer min, @Nullable Integer max) {
+		boolean caseSensitive = Arrays.stream(flags).noneMatch(it -> it == FLAG_CASE_INSENSITIVE);
 		ExecutorService executor = Executors.newSingleThreadExecutor();
 
 		try {
-			RgxGen rgxGen = generateRgxGen(regex, flags);
+			RgxGen rgxGen = generateRgxGen(regex, caseSensitive);
 
 			int minLength = min == null ? 0 : min;
 			int maxLength = max == null ? Integer.MAX_VALUE : max;
@@ -59,25 +58,25 @@ public final class RegexGenerator {
 			List<String> result = new ArrayList<>();
 
 			Future<?> future = executor.submit(() -> {
-				StringIterator stringIterator = rgxGen.iterateUnique();
-				Spliterator<String> spliterator =
-					Spliterators.spliteratorUnknownSize(stringIterator, Spliterator.ORDERED);
-
-				result.addAll(StreamSupport.stream(spliterator, false)
-					.filter(it -> it.length() >= minLength && it.length() <= maxLength)
-					.limit(DEFAULT_REGEXP_GENERATION_MAX_SIZE)
-					.collect(Collectors.toList()));
+				result.addAll(
+					rgxGen.stream()
+						.filter(it -> it.length() >= minLength && it.length() <= maxLength)
+						.limit(DEFAULT_REGEXP_GENERATION_MAX_SIZE)
+						.collect(Collectors.toList())
+				);
 			});
 
 			future.get(DEFAULT_REGEXP_GENERATION_TIMEOUT_SEC, TimeUnit.SECONDS);
 
-			Collections.shuffle(result);
+			List<String> validResults = getValidResults(regex, result, caseSensitive);
+			Collections.shuffle(validResults);
 			return result;
 		} catch (Exception ex) {
 			throw new IllegalArgumentException(
 				String.format(
 					"String generation failed for the regular expression \"%s\" provided in @Pattern."
-						+ " Either the regular expression is incorrect, or it takes too much time to generate",
+						+ " Either the regular expression is incorrect,"
+						+ " or cannot produce a string that matches the regular expression.",
 					regex
 				)
 			);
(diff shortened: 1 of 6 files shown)


fixture-monkey/src/main/java/com/navercorp/fixturemonkey/resolver/ResolvedCombinableArbitrary.java
```diff
@@ -115,7 +115,7 @@ public Object rawValue() {
 		throw new IllegalArgumentException(
 			String.format(
 				"Given type %s could not be generated."
-					+ "Check the ArbitraryIntrospector used or the APIs used in the ArbitraryBuilder.",
+					+ " Check the ArbitraryIntrospector used or the APIs used in the ArbitraryBuilder.",
```
SooKim1110, line 118: Not related to this PR but found some typos.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
```diff
@@ -18,80 +18,83 @@
...
-	}
-
-	private List<String> generateAll(Generex generex, @Nullable Integer min, @Nullable Integer max) {
-		if (min == null) {
-			min = 0;
+	private static RgxGen generateRgxGen(String regex, int[] flags) {
+		RgxGenProperties properties = new RgxGenProperties();
+		if (Arrays.stream(flags).anyMatch(it -> it == 2)) {
+			RgxGenOption.CASE_INSENSITIVE.setInProperties(properties, true);
 		}
```
SooKim1110, line 95: RgxGen only supports CASE_INSENSITIVE flag (https://github.com/curious-odd-man/RgxGen/issues/57).

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
```diff
@@ -18,80 +18,83 @@
...
+		try {
+			RgxGen rgxGen = generateRgxGen(regex, flags);
+
+			int minLength = min == null ? 0 : min;
+			int maxLength = max == null ? Integer.MAX_VALUE : max;
+
+			List<String> result = new ArrayList<>();
+
+			Future<?> future = executor.submit(() -> {
+				StringIterator stringIterator = rgxGen.iterateUnique();
```
SooKim1110, line 64: It might be better to use `generate` than `iterateUnique` because `iterateUnique` produces values in a specific order, resulting in similar generated values.
(It has a downfall of not being able to generate values like `^(?=.*[a-zA-Z])(?=.*[0-9])(?=.*[!@#$%^&*])[a-zA-Z0-9!@#$%^&*]{8,20}$`)

fixture-monkey-api/src/test/java/com/navercorp/fixturemonkey/api/random/RegexGeneratorTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import static org.assertj.core.api.BDDAssertions.then;
+import static org.assertj.core.api.BDDAssertions.thenThrownBy;
+
+import java.util.List;
+import java.util.regex.Pattern;
+
+import org.junit.jupiter.api.Test;
+
+public class RegexGeneratorTest {
```
seongahjo, line 11: `public` is not required.

seongahjo: it may be better to move to the `kotlin-tests` module.
It is more readable thanks to local class.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
```diff
@@ -18,80 +18,83 @@
...
-	public List<String> generateAll(String regex, @Nullable Integer min, @Nullable Integer max) {
-		return this.generateAll(new Generex(regex), min, max);
-	}
-
-	private List<String> generateAll(Generex generex, @Nullable Integer min, @Nullable Integer max) {
-		if (min == null) {
-			min = 0;
+	private static RgxGen generateRgxGen(String regex, int[] flags) {
+		RgxGenProperties properties = new RgxGenProperties();
+		if (Arrays.stream(flags).anyMatch(it -> it == 2)) {
```
seongahjo, line 93: It is better to use static final variable, instead of using magic number.

fixture-monkey-api/src/test/java/com/navercorp/fixturemonkey/api/random/RegexGeneratorTest.java
SooKim1110: I thought it would be better to have the test in the same module as the `RegexGenerator`. Do you prefer moving it?

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
SooKim1110: Changed to use `generate`! Added a regex pattern match inside to prevent fixture generation failing at the `ArbitaryValidator`.

fixture-monkey-api/src/test/java/com/navercorp/fixturemonkey/api/random/RegexGeneratorTest.java
seongahjo: Yes, it is better to test as a client.

Tests in the same module may have a access to a package-private method or class that is not intended.

seongahjo: You are right, in this case, tests in the same module would be better.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
```diff
@@ -18,80 +18,97 @@
...
-		characterClasses.put("\\\\D", "[^0-9]");
-		characterClasses.put("\\\\s", "[ \t\n\f\r]");
-		characterClasses.put("\\\\S", "[^ \t\n\f\r]");
-		characterClasses.put("\\\\w", "[a-zA-Z_0-9]");
-		characterClasses.put("\\\\W", "[^a-zA-Z_0-9]");
-		PREDEFINED_CHARACTER_CLASSES = Collections.unmodifiableMap(characterClasses);
-	}
+	private static final int FLAG_CASE_INSENSITIVE = 2;
+	public static final int DEFAULT_REGEXP_GENERATION_TIMEOUT_SEC = 10;
+	public static final int DEFAULT_REGEXP_GENERATION_MAX_SIZE = 100;
```
seongahjo, line 46: The `public' modifier should be in the first position.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
```diff
@@ -18,80 +18,97 @@
...
-		if (flags.length == 0) {
-			regExp = new RegExp(regex);
-		} else {
-			int intFlag = 0;
-			for (int flag : flags) {
-				intFlag = intFlag | flag;
-			}
-			regExp = new RegExp(regex, intFlag);
+		boolean caseSensitive = Arrays.stream(flags).noneMatch(it -> it == FLAG_CASE_INSENSITIVE);
+		ExecutorService executor = Executors.newSingleThreadExecutor();
```
seongahjo, line 50: `ExecutorService` should be a field of `RegexGenerator`.
There is no need to instantiate it each time.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
```diff
@@ -18,80 +18,97 @@
...
-		characterClasses.put("\\\\S", "[^ \t\n\f\r]");
-		characterClasses.put("\\\\w", "[a-zA-Z_0-9]");
-		characterClasses.put("\\\\W", "[^a-zA-Z_0-9]");
-		PREDEFINED_CHARACTER_CLASSES = Collections.unmodifiableMap(characterClasses);
-	}
+	private static final int FLAG_CASE_INSENSITIVE = 2;
+	public static final int DEFAULT_REGEXP_GENERATION_TIMEOUT_SEC = 10;
+	public static final int DEFAULT_REGEXP_GENERATION_MAX_SIZE = 100;
 
 	public List<String> generateAll(String regex, int[] flags, @Nullable Integer min, @Nullable Integer max) {
```
seongahjo, line 48: I think it is better to deprecate `generateAll`.
It leads to performance degradation.
New API required for the following situation.

https://github.com/naver/fixture-monkey/blob/22fcdadc701224c88281fb7e2858793c049bb992/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/jqwik/JqwikJavaArbitraryResolver.java#L84-L84

SooKim1110: I agree that performance degradation might be an issue here.

Do you have a specific direction in mind? I'm curious whether you are considering a direction that completely avoids supporting pattern validation

flex-myeonghyeon (approved):

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
SooKim1110: actually does variables should be private. Fixed it!

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
SooKim1110: Deprecated `generateAll` and made a method `generate` that only generates one string. It is called lazily with `Arbitraries.ofSuppliers()`.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
```diff
@@ -49,6 +63,38 @@ public final class RegexGenerator {
 		PREDEFINED_CHARACTER_CLASSES = Collections.unmodifiableMap(characterClasses);
 	}
 
+	public String generate(String regex, int[] flags, @Nullable Integer min, @Nullable Integer max, boolean notBlank) {
```
seongahjo, line 66: Since it is a utility method, its parameters should not be too detailed.
It is better to replace `min`, `max`, `notBlank` with `Predicate<String>`

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
```diff
@@ -49,6 +63,38 @@ public final class RegexGenerator {
...
+			RgxGen rgxGen = generateRgxGen(regex, caseSensitive);
+
+			int minLength = min == null ? 0 : min;
+			int maxLength = max == null ? Integer.MAX_VALUE : max;
+
+			String result = executor.submit(() ->
+				rgxGen.stream()
+					.filter(it -> it.length() >= minLength && it.length() <= maxLength)
+					.filter(it -> !(notBlank && it.trim().isEmpty()))
+					.limit(DEFAULT_REGEXP_GENERATION_MAX_SIZE)
```
seongahjo, line 79: Is there a reason why `limit` exists?

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
```diff
@@ -49,6 +63,38 @@ public final class RegexGenerator {
...
+					.findFirst()
+			).get(DEFAULT_REGEXP_GENERATION_TIMEOUT_SEC, TimeUnit.SECONDS).orElse(null);
+
+			checkValidity(regex, result, caseSensitive);
+			return result;
+		} catch (Exception ex) {
+			throw new IllegalArgumentException(
+				String.format(
+					"String generation failed for the regular expression \"%s\" provided in @Pattern."
```
seongahjo, line 88: You don't need to specify that you used `@Pattern`, `RegexGenerator ` is not just for `@Pattern`

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
SooKim1110: Using Predicate<String> seems much better! Thanks for the review.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
SooKim1110: I initially thought that we had to restrict the number of strings generated to be safe, but it appears to be unnecessary. Removed it!

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/RegexGenerator.java
```diff
@@ -94,4 +137,32 @@ private List<String> generateAll(Generex generex, @Nullable Integer min, @Nullab
 		Collections.shuffle(result);
 		return result;
 	}
+
+	private static RgxGen generateRgxGen(String regex, boolean caseSensitive) {
+		RgxGenProperties properties = new RgxGenProperties();
+		if (!caseSensitive) {
+			RgxGenOption.CASE_INSENSITIVE.setInProperties(properties, true);
+		}
```
seongahjo, line 145: It is better to use `caseInsensitive` instead of `!caseSensitive`

seongahjo (approved):
👍

PR #877, opened 2024-01-15
Add handling when using invalid ArbitraryIntrospector

seongahjo:
## Summary
Add logging when using invalid ArbitraryIntrospector

## (Optional): Description
* logs when container type.
* throws an exception if the concrete type.

## How Has This Been Tested?
* existing tests

## Is the Document updated?
No

### DIFF
--- a/docs/content/v1.0.x/release-notes/_index.md
+++ b/docs/content/v1.0.x/release-notes/_index.md
@@ -11,7 +11,7 @@ Fix setting object field by any other type.
 
 Refactor does not throw exception when it cannot generate, the next ArbitraryIntrospector will be used.
 
-Add handling when using invalid ArbitraryIntrospector. Logs when container type, throws an exception if the concrete type.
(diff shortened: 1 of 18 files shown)


docs/content/v1.0.x/release-notes/_index.md
```diff
@@ -5,6 +5,13 @@ menu:
 docs:
 weight: 100
 ---
+sectionStart
+### v1.0.10
+
+Add handling when using invalid ArbitraryIntrospector. Logs when container type, throws an exception if the concrete type.
```
SooKim1110, line 11: A detailed release-note might be better for the users!

Add exception handling when using an `ArbitraryIntrospector` that does not match the property type. Add a log entry if the property is of a container type, and throw an exception if the property is of a concrete type.

SooKim1110 (approved):

flex-myeonghyeon (approved):

PR #934, opened 2024-02-26
feat: 'InnerSpec.keys()' Method with Collection Parameter Support

### LINKED ISSUE #679: 'InnerSpec.keys()' Method with Collection Parameter Support
It would be convenient to have an additional InnerSpec.keys() method that accepts a collection as a parameter.
(same for `values()` and `entries()`)

```kotlin
// As-Is
val keyList = listOf("key1", "key2", "key3")
InnerSpec().property("map") {
    it.keys("key1", "key2").size(2)
}

// To-Be
val keyList = listOf("key1", "key2")
InnerSpec().property("map") {
    it.keys(keyList).size(2)
}
```

related to https://github.com/naver/fixture-monkey/issues/602

No. I would appreciate assistance from a new contributor to add this feature!

songkg7:
## Summary

Supports collection type as a parameter to InnerSpec.keys().

related: resolve #679 

## (Optional): Description

add method as follow:

```java
public InnerSpec keys(Collection<?> keys) {
	keys.forEach(this::key())
	return this;
}
```

However, since the keys() method currently takes an Object as a parameter, I think we should approach method overloading very carefully.

## How Has This Been Tested?

I written test code that takes a list as a parameter.

## Is the Document updated?

No

### DIFF
--- a/fixture-monkey/src/main/java/com/navercorp/fixturemonkey/customizer/InnerSpec.java
+++ b/fixture-monkey/src/main/java/com/navercorp/fixturemonkey/customizer/InnerSpec.java
@@ -24,6 +24,7 @@ import static com.navercorp.fixturemonkey.Constants.NO_OR_ALL_INDEX_INTEGER_VALU
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
@@ -163,6 +164,16 @@ public final class InnerSpec {
 	public InnerSpec key(Object key) {
 		entrySize++;
 		setMapKey(key);
+		return this;
+	}
+
+	/**
+	 * Sets Collection of keys in the currently referred map property.
+	 *
+	 * @param keys The Collection of keys to set in the map. Can be empty.
+	 */
+	public InnerSpec keys(Collection<?> keys) {
+		keys.forEach(this::key);
 		return this;
 	}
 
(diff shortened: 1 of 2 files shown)


fixture-monkey/src/main/java/com/navercorp/fixturemonkey/customizer/InnerSpec.java
```diff
@@ -166,6 +167,16 @@ public InnerSpec key(Object key) {
 		return this;
 	}
 
+	/**
+	 * Sets Collection of keys in the currently referred map property.
```
SooKim1110, line 171: `Sets multiple keys in the currently referred map property from a Collection.`
로 주석이 달리면 조금 더 명확할 것 같습니다!

SooKim1110:
Thanks for your contribution! 
If you don't mind, it would be great to have a similar method that accepts a collection as a parameter for values() and entries().

fixture-monkey/src/test/java/com/navercorp/fixturemonkey/test/InnerSpecTest.java
```diff
@@ -827,4 +828,18 @@ void setNotNull() {
 
 		then(actual).isNotNull();
 	}
+
+	@Property
+	void keysForCollection() {
+		List<String> keyList = Arrays.asList("key1", "key2", "key3");
+		Map<String, String> actual = SUT.giveMeBuilder(MapObject.class)
```
SooKim1110, line 835: It would be nice to add a blank line in between so that the given-when-then pattern is more visible in the test.

fixture-monkey/src/test/java/com/navercorp/fixturemonkey/test/InnerSpecTest.java
```diff
@@ -827,4 +828,18 @@ void setNotNull() {
...
+			.setInner(
+				new InnerSpec()
+					.property("strMap", it -> it.keys(keyList).size(3))
+			)
+			.sample()
+			.getStrMap();
+
+		then(actual.keySet()).contains("key1", "key2", "key3");
```
SooKim1110, line 843: You can also use `containsAll' instead of `contains`, which makes the code shorter.

PR #950, opened 2024-03-16
Module rearange and add kotlin benchmark

jinia91:
## Summary

I suggest some modifications to the current structure to facilitate benchmark testing of the `fixture-monkey-kotlin`module using JMH

1. To ensure consistent results, use the same class and use the `java-test-fixtures` plugin for code reuse.

2. make new module `fixture-monkey-benchmarks` for aggregating sub-benchmark-module.

2. Due to some local compile issues when defining Java classes within the Kotlin module, move the `ExpressionGeneratorJavaTestSpecs.java` to a Java package in `fixture-monkey-kotlin`.

3. Add default benchmark samples to the `fixture-monkey-kotlin` module.

        For a proper comparison, the sample test cases are planned as follows:
        
        1. Benchmark for initializing Java classes in Kotlin
        2. Benchmark for initializing Kotlin classes in Kotlin
        3. Benchmark for initializing composite classes with Kotlin-Java internal references (todo)

5. Modify the GitHub Actions workflow for benchmark reporting.

## How Has This Been Tested?
Local Build and Ci Pipline

## Is the Document updated?
no need


seongahjo:
How about adding a new module to benchmark?

jinia91:
> How about adding a new module to benchmark?

Im considering two approaches 

- first, make new module for benchmark integration testing
- second, performing benchmark testing for each module separately and collect just reporting. 

From the perspective of scalability, choice second approach.

However, I'm not entirely confident due to the downside of increased dependency complexity.

jinia91:
| Java Benchmark                                                                                                        | Mode | Threads | Samples | Score       | Score Error (99.9%) | Unit  |
| ---------------------------------------------------------------------------------------------------------------- | ---- | ------- | ------- | ----------- | ------------------- | ----- |
| com.navercorp.fixturemonkey.ManipulationBenchmark.fixed                                                          | avgt | 1       | 10      | 576.165762  | 2.457355            | ms/op |
| com.navercorp.fixturemonkey.ManipulationBenchmark.thenApply                                                      | avgt | 1       | 10      | 1301.278260 | 5.921[34](https://github.com/naver/fixture-monkey/actions/runs/8306048573/job/22733656974?pr=950#step:6:35)3            | ms/op |
| com.navercorp.fixturemonkey.ObjectGenerationBenchmark.beanGenerateOrderSheetWithFixtureMonkey                    | avgt | 1       | 10      | 596.718850  | 5.575784            | ms/op |
| com.navercorp.fixturemonkey.ObjectGenerationBenchmark.builderGenerateOrderSheetWithFixtureMonkey                 | avgt | 1       | 10      | 505.[36](https://github.com/naver/fixture-monkey/actions/runs/8306048573/job/22733656974?pr=950#step:6:37)5829  | 3.092013            | ms/op |
| com.navercorp.fixturemonkey.ObjectGenerationBenchmark.fieldReflectionGenerateOrderSheetWithFixtureMonkey         | avgt | 1       | 10      | 572.670919  | 6.588839            | ms/op |
| com.navercorp.fixturemonkey.ObjectGenerationBenchmark.jacksonGenerateOrderSheetWithFixtureMonkey                 | avgt | 1       | 10      | 607.734411  | 4.826412            | ms/op |
|                                                                                                                  |
| Kotlin Module Benchmark                                                                                                        | Mode | Threads | Samples | Score       | Score Error (99.9%) | Unit  |
| com.navercorp.fixturemonkey.kotlin.KotlinObjectGenerationBenchMark.beanGenerateJavaOrderSheetWithFixtureMonkey   | avgt | 1       | 10      | 1029.619206 | 21.428195           | ms/op |
| com.navercorp.fixturemonkey.kotlin.KotlinObjectGenerationBenchMark.beanGenerateKotlinOrderSheetWithFixtureMonkey | avgt | 1       | 10      | 980.861308  | 14.2[37](https://github.com/naver/fixture-monkey/actions/runs/8306048573/job/22733656974?pr=950#step:6:38)564           | ms/op |

seongahjo:
I'm sure you've given it a lot of thought. 
Can you tell me about a specific situation where flexibility is critical?
I have no idea so far. I'm not sure about it is worth than complex dependency.

jinia91:
Currently, the execution speed of performance bench check feels too slow, leading to a decrease in productivity. 

With the current PR, as adding just two of kotlin benchmarking, there has been an increase of over 4 minutes. 

As fixture-monkey continues to expand with more tests being added, i expect to take more times in future.

so, in near future, try to various pipelines such as:

- Running only some benchmarks based on module changes
- Executing each benchmark in parallel
and so on...etc..

are expected to be considered. 

In such cases, big one module for benchmark might lead to break changes.

Of course, this is just speculative in my opinion, and the problems of increased dependency complexity and management points cannot be ignored.

seongahjo:
Oh, I see, thank you for your detailed explanation. Yes, I agree that a large module leads to bad for performance, which is bad for productivity. then how about a new module that contains benchmark modules like `fixture-monkey-tests`?

jinia91:
Oh, that sounds good.

There seems to be no need to place the benchmark testing module under the basic module structure unnecessarily.
I will change the direction to manage under the `fixture-monkey-tests` each benchmark modules.

Thank you for the great idea.

fixture-monkey-benchmarks/fixture-monkey-benchmark/src/jmh/java/com/navercorp/fixturemonkey/BuilderOrderSheet.java
```diff
@@ -174,4 +174,3 @@ public enum BundleType {
 		IDENTICAL_PRODUCT
 	}
 }
-
```
seongahjo, line 177: EOL is missing now.

fixture-monkey-benchmarks/fixture-monkey-benchmark/src/testFixtures/java/com/navercorp/fixturemonkey/OrderSheet.java
```diff
@@ -180,4 +180,3 @@ public enum BundleType {
 		IDENTICAL_PRODUCT
 	}
 }
-
```
seongahjo, line 183: EOL is missing now.

fixture-monkey-kotlin/build.gradle.kts
```diff
@@ -10,6 +10,7 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-reflect:${Versions.KOTLIN}")
 
     testImplementation(project(":fixture-monkey-jackson"))
+    testImplementation(project(":fixture-monkey-javax-validation"))
```
seongahjo, line 13: Is this dependency necessary?

fixture-monkey-kotlin/src/test/java/com/navercorp/fixturemonkey/kotlin/spec/ExpressionGeneratorJavaTestSpecs.java
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-package com.navercorp.fixturemonkey.kotlin;
+package com.navercorp.fixturemonkey.kotlin.spec;
```
seongahjo, line 19: I think this change has nothing to do with this PR.

fixture-monkey-kotlin/src/test/kotlin/com/navercorp/fixturemonkey/kotlin/ExpressionGeneratorsTest.kt
```diff
@@ -18,8 +18,8 @@
 
 package com.navercorp.fixturemonkey.kotlin
 
-import com.navercorp.fixturemonkey.kotlin.ExpressionGeneratorJavaTestSpecs.DogJava
-import com.navercorp.fixturemonkey.kotlin.ExpressionGeneratorJavaTestSpecs.PersonJava
+import com.navercorp.fixturemonkey.kotlin.spec.ExpressionGeneratorJavaTestSpecs.DogJava
+import com.navercorp.fixturemonkey.kotlin.spec.ExpressionGeneratorJavaTestSpecs.PersonJava
```
seongahjo, line 22: I think this change has nothing to do with this PR.

seongahjo (approved):
Only minor convention changes are required.
Nice work 👍

fixture-monkey-kotlin/build.gradle.kts
jinia91: In the ExpressionGeneratorJavaTestSpecs, javax validation was previously being used. 

Even on the main branch, compile was not successful in my local environment, the exact reason is unclear.

After changing modules, [compile fails in the absence of the dependency](https://github.com/naver/fixture-monkey/actions/runs/8321197148/job/22767082469), so it necessary in the current situation.

Issue #678, opened 2023-06-16
support generating lambda type


seongahjo:
## Objective
support generating lambda type.

### Java
* Supplier<T>
* Function<T, U>

### Kotlin
* [Function](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-function.html)

### Requirements
1. could be set lambda itself, for example `set("supplier", (Supplier<String>) () -> "test")`

mollder:
Hi, I'm interested in this open source, and if this issue is not reflected yet, is it okay for me to try to solve this issue?

seongahjo:
@mollder 
Hello, sure. Thank you for your interest. 
If you have any questions, please do not hesitate to ask.

I think it is good approach to take a look at the implementations of`ArbitraryIntrospector` and `PropertyGenerator`

mollder:
Thank you for the explanation.

I'll look at the Arbitrary Introspector and Property Generator that you mentioned, and then I'll aim to raise the pr in about two weeks

PR #951, opened 2024-03-16
support java supplier type

mollder:
## Summary
https://github.com/naver/fixture-monkey/issues/678

Originally, I was going to provide Java Fuction types and Kotlin function types, but as I implemented the Supplier, I had some questions (in NodeSetLazyManipulator class), and I thought it would be better to talk about this first and then implement other types, so I posted the Supplier first.

The questions I have are as follows.

1 operate the set code below, the Manipulator does not return the Supplier, but returns the value inside it. to solve this problem i forcibly wrapped it with the Supplier once more in the NodeSetLazyManipulator.

`set("supplier", (Supplier<String>) () -> "test")`

Although my code solve issue, it is so hard to recognize and I think it is easy to cause bugs in long-term perspective, so I was wondering what you think about fixing the feature as below.

- What do you think about adding an setLazy method to the user?
- what do you think about removing behavior of returning the value Lazyly When delivering Supplier parameters to the set method?

2 in CombinableArbitrary, it is supposed to return the value in a Lazy way once again, and I tried to use FixedCombinableArbitrary to return the Supplier itself, but it was blocked by dependence. So in order to solve issue, I forced to wrap it up with a Supplier once more. i think it is not good

- What do you think about adding the fromLazy method from the Util methods of CombinableArbitrary and removing it from the from (Supplier value) method?

3 i haven't implemented InnerSpec yet, do you want to support InnerSpec?
(When I tested it, I found that innerSpec did not work well yet with Supplier)

## How Has This Been Tested?
Multiple tests have been added in the FixureMonkeyTest class.

## Is the Document updated?
No

I think it's better to add documentation after adding other types, but I wonder what you think about this.

If you want, I'll add the documentation to the location you want.


fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/introspector/SupplierIntrospector.java
```diff
@@ -0,0 +1,57 @@
...
+import com.navercorp.fixturemonkey.api.matcher.AssignableTypeMatcher;
+import com.navercorp.fixturemonkey.api.matcher.Matcher;
+import com.navercorp.fixturemonkey.api.property.Property;
+
+import java.util.function.Supplier;
+
+import static com.navercorp.fixturemonkey.api.matcher.SingleGenericTypeMatcher.SINGLE_GENERIC_TYPE_MATCHER;
+
+public final class SupplierIntrospector implements ArbitraryIntrospector, Matcher {
+	private static final Matcher MATCHER = new AssignableTypeMatcher(Supplier.class);
```
seongahjo, line 33: Using `intersect` would be more clearer, instead of `SINGLE_GENERIC_TYPE_MATCHER.match(property) && MATCHER.match(property)`

seongahjo:
@mollder 
Thank you for your contributing 😃 

I'll answer your questions later comments.
Before that, I want to know if your approach can handle this situation.

```java
class Foo {
  String foo;
}

String actual = fixture.giveMeBuilder(new TypeReference<Supplier<Foo>>(){})
    .set("$.foo", "expected")
    .sample()
    .get()
     .foo;
     
then(actual).isEqualTo("expected")
```

seongahjo:
>  1 What do you think about adding an setLazy method to the user?
what do you think about removing behavior of returning the value Lazyly When delivering Supplier parameters to the set method?

I think you want to set `Supplier` itself. In order to do that, You must wrapping it by `Values.just`.
Just like below.

`set("supplier", Values.just((Supplier<String>)() -> "test")))`


> 2 in CombinableArbitrary, it is supposed to return the value in a Lazy way once again, and I tried to use FixedCombinableArbitrary to return the Supplier itself, but it was blocked by dependence. So in order to solve issue, I forced to wrap it up with a Supplier once more. i think it is not good
What do you think about adding the fromLazy method from the Util methods of CombinableArbitrary and removing it from the from (Supplier value) method?

I see, but it should be careful to add new method `fromLazy` and remove `form(Supplier)`, because it leads to breaking change.
I think it is better to support `Values.just` in `CombinableArbitrary` just like `set`.

> 3 i haven't implemented InnerSpec yet, do you want to support InnerSpec?
(When I tested it, I found that innerSpec did not work well yet with Supplier)

Can you tell me more details about it?

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/introspector/SupplierIntrospector.java
mollder: [f39b153](https://github.com/naver/fixture-monkey/pull/951/commits/f39b153e9b0eb236b6b70cf5159a4e11c4481d59)

mollder:
@seongahjo 

Thank you for your detailed answer 👍 👍 

1 using Value.just

> I think you want to set Supplier itself. In order to do that, You must wrapping it by Values.just.
> Just like below.
> 
> set("supplier", Values.just((Supplier<String>)() -> "test")))

> I see, but it should be careful to add new method fromLazy and remove form(Supplier), because it leads to breaking change.
> I think it is better to support Values.just in CombinableArbitrary just like set.

I didn't know about how to use Value.just, and I changed it because I thought it was better to use Value.just

[6fdb1b0](https://github.com/naver/fixture-monkey/pull/951/commits/6fdb1b0aa17703086bea536b3db9467a65cf004d)

2 Expression bug

```
class Foo {
  String foo;
}

String actual = fixture.giveMeBuilder(new TypeReference<Supplier<Foo>>(){})
    .set("$.foo", "expected")
    .sample()
    .get()
     .foo;
     
then(actual).isEqualTo("expected")
```

As a result of my debugging, this bug was occurring because rootNode couldn't get children when I was covering objects like Suppliers or Optional.

So I fixed the problem by changing the IdentityNodeResolver to point directly to the child node.

how do you think about this approach

[af05fd2](https://github.com/naver/fixture-monkey/pull/951/commits/af05fd290e9207cac52210a0e694411b335b2878)

3 inner spec

I think if using Supplier in innerSpec, it won't work, so I asked the question
But I actually wrote a test using Value.just and it worked well, so it is just my misunderstood

[37786db](https://github.com/naver/fixture-monkey/pull/951/commits/37786db9ec42c57bcc68516b0ca119c8723916a3)

seongahjo:
@mollder 
> So I fixed the problem by changing the IdentityNodeResolver to point directly to the child node. how do you think about this approach

Actually, `IdentityNodeResolver` is a node resolver that is converted from `$` (root).
It should not know the properties of the child because it is not its responsibility.

For example, `$.str` is converted to a combination of `IdentityNodeResolver` and `PropertyNameResolver`.

I think the reason it does not work without changing `IdentityNodeResolver` is that the `Supplier` type is treated as a container. The elements of the container type are treated as `ElementProperty' which has no property name. In order to referencing `ElementProperty`, you should use the expression `[0]`.

So, I think it is time to create a new property for type like `Optional`, `Supplier` etc.

What do you think?

mollder:
@seongahjo 

sorry to late reply

Are you thinking of creating a new Expression like (0) like the code below?

I thought there should be a new expression to create a CompositeNodeResolver in the toNodeResolver method of the ArbitraryExpression class.

And it becomes easier when adding a Function type or a BiFunction type

```
class Foo {
  String foo;
}

String actual = fixture.giveMeBuilder(new TypeReference<Supplier<Foo>>(){})
    .set("$(0).foo", "expected")
    .sample()
    .get()
     .foo;
```

mollder:
If you have thought about not to make a new Expression, please comment

seongahjo:
@mollder 
>  If you have thought about not to make a new Expression, please comment

```java
String actual = fixture.giveMeBuilder(new TypeReference<Supplier<Foo>>(){})
    .set("$.foo", "expected")
    .sample()
    .get()
     .foo;
```

No, I think it should be used the expression as above.
It should work without modifying `IdentityNodeResolver`

I think it could be done by using `ContainerPropertyGenerator`

mollder:
ok I'll try it in that direction

Thank you for commenting every time

mollder:
@seongahjo 

I think the ObjectNode consists of three stages, including children and grandchildren, is scalable structure when SupplierIntrospector generates values, so I left that structure intact

However, if the ObjectNode consists of three stages, it is impossible to set the expression with the existing NodeResolver, so I added a WrappedNodeResolver and modified the test breaks accordingly.

What do you think of this direction?

mollder:
@seongahjo [69353ab](https://github.com/naver/fixture-monkey/pull/951/commits/69353ab65e88ed82cab2d44f7b326c5bea6d0c2d)

As i understood, I have modified the code
Is this the right direction you comment?

seongahjo:
@mollder 
Yes, that is exactly what I am thinking of.

Only one small change is needed.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/generator/OptionalContainerPropertyGenerator.java
```diff
@@ -49,15 +54,42 @@ public ContainerProperty generate(ContainerPropertyGeneratorContext context) {
...
 		AnnotatedType valueAnnotatedType = getValueAnnotatedType(property);
-		Property valueProperty = new ElementProperty(
-			property,
-			valueAnnotatedType,
-			0,
-			0
-		);
+		Type childType = valueAnnotatedType.getType();
+		AnnotatedType childAnnotatedType = Types.generateAnnotatedTypeWithoutAnnotation(childType);
+
```
seongahjo, line 59: Is there a difference between `valueAnnotatedType` and `childAnnotatedType`?

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/property/SingleElementProperty.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.OptionalLong;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+import javax.annotation.Nullable;
+
+import com.navercorp.fixturemonkey.api.type.Types;
+
+public final class SingleElementProperty implements Property {
```
seongahjo, line 40: `@API(since = "1.0.16", status = Status.EXPERIMENTAL)` is needed.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/property/SingleElementProperty.java
```diff
@@ -0,0 +1,155 @@
...
+		if (isOptional(actualType)) {
+			return getOptionalValue(instance);
+		}
+
+		if (Supplier.class.isAssignableFrom(actualType)) {
+			return instance;
+		}
+
+		throw new IllegalArgumentException("given value has no match");
+	}
```
seongahjo, line 106: It is better to delegate to the`property` field.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/generator/OptionalContainerPropertyGenerator.java
mollder: [0ce0cb1](https://github.com/naver/fixture-monkey/pull/951/commits/0ce0cb1e5c282bf4bd8a912b1995fb353720b1fd)
[673171a](https://github.com/naver/fixture-monkey/pull/951/commits/673171a6714f2645abfcf5cb6f44cebddc7dfb20)

I want Supplier<SimpleObject> type to property annotateType in the Supplier<Supplier<SimpleObject>> case, but I was mistaken and deleted the childAnnotateType

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/property/SingleElementProperty.java
mollder: [9edca18](https://github.com/naver/fixture-monkey/pull/951/commits/9edca1840cf2c7c1bf404bf8224e563c604b40c6)
[c827311](https://github.com/naver/fixture-monkey/pull/951/commits/c82731111a1f39edee8f5fe627d7e0247eef8a6e)

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/property/SingleElementProperty.java
mollder: [a1f6314](https://github.com/naver/fixture-monkey/pull/951/commits/a1f6314a13d180d9ae532dee9233fc6ceb76ecc7)
[72103fd](https://github.com/naver/fixture-monkey/pull/951/commits/72103fd87cb7b910b424f285e627309566e83856)

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/generator/OptionalContainerPropertyGenerator.java
```diff
@@ -48,21 +50,19 @@ public final class OptionalContainerPropertyGenerator implements ContainerProper
...
-			property,
-			valueAnnotatedType,
-			0,
-			0
-		);
+		AnnotatedType valueAnnotatedType = getOptionalValueAnnotatedType(property);
+		Type valueType = valueAnnotatedType.getType();
+
+		Property childProperty = new OptionalProperty(valueType, valueAnnotatedType);
+		SingleElementProperty singleElementProperty = new SingleElementProperty(childProperty, valueAnnotatedType);
```
seongahjo, line 57: I think it could be enough for `SingleElementProperty`.
Is it not work if only one `SingleElementProperty` like below?

```java
AnnotatedType valueAnnotatedType = getOptionalValueAnnotatedType(property);
		Type valueType = valueAnnotatedType.getType();

SingleElementProperty singleElementProperty = new SingleElementProperty(valueAnnotatedType);
```

I do not know why `SingleElementProperty` needs `valueAnnotatedType' as it is the annotated type of `childProperty'.

I think I only need one or the other between `childProperty` and `valueAnnotatedType`.

I'm not sure it's right time to split into `OptionalProperty` and `SupplierProperty`.

mollder: > I think it could be enough for SingleElementProperty.
> Is it not work if only one SingleElementProperty like below?
> 
> ```
> AnnotatedType valueAnnotatedType = getOptionalValueAnnotatedType(property);
> 		Type valueType = valueAnnotatedType.getType();
> 
> SingleElementProperty singleElementProperty = new SingleElementProperty(valueAnnotatedType);
> ```
> 
> I do not know why SingleElementProperty needs valueAnnotatedType' as it is the annotated type of childProperty'.


if SingleElementProperty(remove property field) like below, test broken and i think resson is cannot get (Supplier, Optional)'s internal value

```
AnnotatedType valueAnnotatedType = getOptionalValueAnnotatedType(property);

SingleElementProperty singleElementProperty = new SingleElementProperty(valueAnnotatedType);
```

and if only childProperty, is same behavior previous commit. work well

so i remove valueAnnotatedType parameter to SingleElementProperty constructor

> I'm not sure it's right time to split into OptionalProperty and SupplierProperty.

and I respect your comment, and I didn't separate it with confidence, only the feeling that role separation would be better, so i rollback to anonymous class

[8173376](https://github.com/naver/fixture-monkey/pull/951/commits/81733767bdc2bf552281f5c24b1d1f4b8d5faa42)

seongahjo (approved):
👍  It is a difficult job, but you have done it perfectly.
Thank you for your effort.

Could you resolve the conflict?

mollder:
conflict resolved

and thank for your comment 👍 
i can't do it without your comment

Issue #962, opened 2024-03-24
I want fields to be injected with a Kotlin Duration type


this-is-spear:
## Describe the feature you request
The user doesn't understand why the field needs to be injected with the `Long` type.

``` kotlin
sut.giveMeBuilder<DurationValue>()
    .set(DurationValue::duration, Long.MAX_VALUE)
    .sample()
```

## Suggest A Solution

For ease of use, Duration needs to be modified to allow you to inject Duration values.

``` kotlin
sut.giveMeBuilder<DurationValue>()
    .set(DurationValue::duration, Duration.INFINITE)
    .sample()
```

## If the feature request is approved, would you be willing to submit a PR?
*Yes*

seongahjo:
@this-is-spear 
Sounds great.  Could you make a PR for it?

this-is-spear:
Okay, good!

PR #967, opened 2024-04-05
Modify the way a value class is used to output arbitrary value

this-is-spear:
## Summary

#962

Fixed to use kotlin duration in the set function.

## (Optional): Description

Fixed `NodeSetDecomposedValueManipulator` throwing an exception because the object node and value class types did not match.

Fixed an issue in `PrimaryConstructorArbitraryIntrospector#resolveArbitrary` where when injecting `Duration`, it was trying to inject Duration class as Duration because Duration is a value.

As we progressed, there was an issue where the value class didn't have its own type and the operation was proceeding.
So we fixed the code so that the value class can continue working with its own type.

## How Has This Been Tested?

To test Duration, we added or reused the validation logic below.

- Tested in `PrimaryConstructorArbitraryIntrospectorTest#setDurationValue`.
- Tested in `PrimaryConstructorArbitraryIntrospectorTest#sampleDurationValue`.
- Tested in `PrimaryConstructorArbitraryIntrospectorTest#sampleDuration`.

To test Value class, we added or reused the validation logic below.

- Tested in `ValueClassTest#foo`
- Tested in `ValueClassTest#valueClassProperty`
- Tested in `ValueClassTest#setValueClassProperty`

## Is the Document updated?

We didn't update the documentation because there were no feature changes.


fixture-monkey-kotlin/src/main/kotlin/com/navercorp/fixturemonkey/kotlin/introspector/PrimaryConstructorArbitraryIntrospector.kt
```diff
@@ -82,7 +83,9 @@ class PrimaryConstructorArbitraryIntrospector : ArbitraryIntrospector {
     ): Any? {
         return try {
             val parameterKotlinType = parameter.type.jvmErasure
-            if (parameterKotlinType.isValue) {
+            if (arbitrariesByPropertyName[parameter.name] is Duration) {
+                arbitrariesByPropertyName[parameter.name]
+            } else if (parameterKotlinType.isValue) {
```
seongahjo, line 88: I think it is a better approach to modify `KotlinPropertyCache#getMemberProperties` for `value class`.

this-is-spear: There's no case for constructor injection for the value class, because it's injected directly into the value.

So I don't think we need to do anything to inject constructors.

this-is-spear: It's not working the way we thought it would! We'll ask for a review again when we're done.

seongahjo: I think it is a good approach to add a new implementation of `Property` for value class, such as`ValueClassProperty`. `KotlinPropertyCache#getMemberProperties` would generate a `ValueClassProperty` for value class.

In this approach, You do not need to modify a specific ArbitraryIntrospector, `PrimaryConstructorArbitraryIntrospector`

All you have to do is add a new ArbitraryIntrospector for `ValueClassProperty`.
Then, you do not need to an `if` statement below because a new ArbitraryIntrospector will generate a value class type  `CombinableArbitrary`.

```kotlin
  if (parameterKotlinType.isValue) {
                parameterKotlinType.primaryConstructor!!.isAccessible = true
                parameterKotlinType.primaryConstructor!!.call(arbitrariesByPropertyName[parameter.name])
            }
```

this-is-spear: I The reason written to handle value class in the PrimaryConstructorArbitraryIntrospector is because the actual type is different from the value returned as AnnotatedType in KPropertyProperty.

For value objects, I modified it to include the existing type and it worked fine.

fixture-monkey-kotlin/src/main/kotlin/com/navercorp/fixturemonkey/kotlin/property/KPropertyProperty.kt
```diff
@@ -44,7 +44,8 @@ data class KPropertyProperty(
     override fun getValue(instance: Any): Any? {
         val javaMethod = this.kProperty.getter.javaMethod
 
-        if (javaMethod != null && Modifier.isPublic(javaMethod.modifiers)) {
+        if (javaMethod != null) {
+            this.kProperty.isAccessible = true
```
seongahjo, line 48: is it necessary for `value class`?

fixture-monkey-kotlin/src/main/kotlin/com/navercorp/fixturemonkey/kotlin/type/KotlinTypes.kt
```diff
@@ -146,3 +147,12 @@ fun Class<*>.declaredConstructor(
...
     } ?: this.kotlin.constructors.first()
+
+fun KProperty<*>.getValueClassType(): Type {
+    val classifier = returnType.classifier
+    return if (classifier is KClass<*> && classifier.isValue) {
+        classifier.java
+    } else {
+        returnType.javaType
+    }
+}
```
seongahjo, line 158: For clarity, I think it is better to inline the method.

The name `getValueClassType` confuses me as it is the method for `value class` only.

fixture-monkey-kotlin/src/main/kotlin/com/navercorp/fixturemonkey/kotlin/property/KPropertyProperty.kt
this-is-spear: Duration has a rawValue for the StorageUnit, so we need to know some private method values to get the StorageUnit information.

```kotlin
private val storageUnit get() = if (isInNanos()) DurationUnit.NANOSECONDS else DurationUnit.MILLISECONDS
```

The get method of the StorageUnit is declared private, which means that it can't be called from outside. The code below is the format that would be added if you decompiled the compiled code from the class file again.

```kotlin
private final val storageUnit: kotlin.time.DurationUnit /* compiled code */
    private final get

```

fixture-monkey-kotlin/src/main/kotlin/com/navercorp/fixturemonkey/kotlin/type/KotlinTypes.kt
this-is-spear: That sounds great!
I remove getValueClassType to inline the caller method. 👍

this-is-spear:
I modified the PR topic because the focus was on the work done to get the Value class creation working correctly. I believe one of the subtasks was to fix the Duration class to work correctly.

I thought it needed to be fixed in order to see and understand the git history.

If you need to change to the previous topic, please let me know. I'll fix it. ☺️

seongahjo (approved):
It is a difficult problem, but you have solved it!

Thank you

this-is-spear:
It's all thanks to the maintainer's help!
I learned a lot 🙇‍♂️

PR #991, opened 2024-06-17
Add support multi-level inheritance of sealed

ctmanjak:
## Summary
Add support for multi-level inheritance of sealed class and sealed interface.

## How Has This Been Tested?
Add a new test that adds an intermediate inherited class and interface to an existing test.

PR #992, opened 2024-06-17
Fix when to apply JdkVariantOptions

ctmanjak:
## Summary
Fix when to apply JdkVariantOptions in FixtureMonkeyOptionsBuilder.

## (Optional): Description
Fix an issue where JdkVariantOptions.apply has a higher priority than options set by the user because it is in the build method of FixtureMonkeyOptionsBuilder.

## How Has This Been Tested?
Use existing tests.

### DIFF
--- a/fixture-monkey-tests/java-17-tests/src/test/java/com/navercorp/fixturemonkey/tests/java17/JdkVariantOptionsTest.java
+++ b/fixture-monkey-tests/java-17-tests/src/test/java/com/navercorp/fixturemonkey/tests/java17/JdkVariantOptionsTest.java
@@ -20,15 +20,15 @@ class JdkVariantOptionsTest {
 	void sampleSealedClass() {
 		// given
 		FixtureMonkey sut = FixtureMonkey.builder()
-		.objectIntrospector(ConstructorPropertiesArbitraryIntrospector.INSTANCE)
-		.defaultNotNull(true)
-		.pushAssignableTypeObjectPropertyGenerator(
-			BaseSealedClass.class,
-			new InterfaceObjectPropertyGenerator<>(
-				List.of(SealedClass.class)
-			)
-		)
-		.build();
+		    .objectIntrospector(ConstructorPropertiesArbitraryIntrospector.INSTANCE)
+		    .defaultNotNull(true)
+		    .pushAssignableTypeObjectPropertyGenerator(
+			    BaseSealedClass.class,
+			    new InterfaceObjectPropertyGenerator<>(
+			    	List.of(SealedClass.class)
+			    )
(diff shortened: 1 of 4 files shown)


seongahjo:
I know when to modify the `JdkVariantOption`.

Could you show the example?

ctmanjak:
@seongahjo 
```
private static final FixtureMonkey SUT = FixtureMonkey.builder()
	.objectIntrospector(ConstructorPropertiesArbitraryIntrospector.INSTANCE)
	.defaultNotNull(true)
	.pushAssignableTypeObjectPropertyGenerator(
		SealedClass.class,
		new InterfaceObjectPropertyGenerator<>(
			List.of(
				SealedClassImpl.class
			)
		)
	)
	.build();

private static sealed class SealedClass permits SealedClassImpl, SealedClassImpl2 {
}

private static final class SealedClassImpl extends SealedClass {
}

private static final class SealedClassImpl2 extends SealedClass {
}
```
If you declared it in the same way as above.


```
@RepeatedTest(TEST_COUNT)
void sampleSealedClass() {
	// when
	SealedClass actual = SUT.giveMeOne(SealedClass.class);

	then(actual).isInstanceOf(SealedClassImpl.class);
}
```
About half of these tests will fail.

스크린샷 2024-06-17 오후 8 05 31

because the ObjectPropertyGenerator applied by JdkVariantOptions has a higher priority than the one you added.

seongahjo (approved):

seongahjo:
@ctmanjak 
If so, it is better to add the test with the modified option to illustrate the case.

ctmanjak:
@seongahjo
I think I'll need the SealedClassTestSpecs from #991 when writing test code, should I combine the two PRs?

fixture-monkey-tests/java-17-tests/src/test/java/com/navercorp/fixturemonkey/tests/java17/JdkVariantOptionsTest.java
```diff
@@ -0,0 +1,37 @@
...
+				List.of(SealedClass.class)
+			)
+		)
+		.build();
+
+	@RepeatedTest(TEST_COUNT)
+	@DisplayName("User-added ObjectPropertyGenerators have a higher priority.")
+	void sampleSealedClass() {
+		// when
+		BaseSealedClass actual = SUT.giveMeOne(BaseSealedClass.class);
```
seongahjo, line 33: In this case, I think a `Fixture Monkey` instance should be in the test case.

It is a crucial information that Fixture Monkey use `pushAssignableTypeObjectPropertyGenerator` option.

PR #1067, opened 2024-10-22
Add base methods of Object to proxy object of interface

acktsap:
## Summary

Add basic method of java Object

- hashCode
- equals
- toString

## (Optional): Description

Without these methods, it fails in shrinking process when instance is generated using interface.

## How Has This Been Tested?

Yes

## Is the Document updated?

Nope


seongahjo (approved):
👍  수고하셨습니다

Issue #1127, opened 2024-12-20
Incorrect minimum value comparison logic in Jakarta/Javax ValidationConstraintGenerator


Seol-JY:
## Describe the bug
In `JakartaValidationConstraintGenerator` class ([line 365](https://github.com/naver/fixture-monkey/blob/b02f51185bd8943acc73ebc146695d205f2ba346/fixture-monkey-jakarta-validation/src/main/java/com/navercorp/fixturemonkey/jakarta/validation/introspector/JakartaValidationConstraintGenerator.java#L365)), there appears to be a meaningless self-comparison operation:

```java
negativeMin = negativeMin.min(negativeMin);
```

It appears this was intended to be:
 ```java
negativeMin = negativeMin.min(decimalMin);
```

For what seems to be a mistake, I tried to verify with the following test case:
```java
public static class A {
    public A() {}

    public A(BigDecimal value) {
        this.value = value;
    }

    @NotNull
    @Max(-49)
    @Min(-100)
    @DecimalMin("-49.9")
    private BigDecimal value;

    public BigDecimal getValue() {
        return value;
    }

    public void setValue(BigDecimal value) {
        this.value = value;
    }
}
```
Expected the final `negativeMin` to be -100 which should cause validation errors, but the code runs without exceptions.

## Your environment
* version of Fixture Monkey: 1.1.5
* version of Java: 17

The code contains a self-comparison operation in its minimum value calculation, yet the validation process operates as expected in test cases.

I would like to understand:
Whether this code represents an actual issue
If it is an issue, why the test case still functions correctly

seongahjo:
@Seol-JY 
Hi!

Thank you for reporting an issue.

> Whether this code represents an actual issue If it is an issue, why the test case still functions correctly

That's because Fixture Monkey has two logical flows to satisfy the validation. [link](https://github.com/naver/fixture-monkey/issues/1118#issuecomment-2535135624)

Although the first flow (JakartaValidationConstraintGenerator) made a mistake, the second flow takes care of the problem.

It seems you are interested in contributing, could you do a PR for this issue?

Thank you.

PR #1131, opened 2024-12-23
Enhance decimal validation to match actual Jakarta Validation behavior and fix related issues

### LINKED ISSUE #1126: IllegalArgumentException when generating non-integer numeric values with equal min and max values
## Describe the bug
When trying to generate test values using FixtureMonkey, an IllegalArgumentException is thrown if the minimum and maximum values are equal and the borders are not included (from jqwik). This issue affects BigDecimal and floating-point types (float, double) when using similar validation constraints.

## Your environment
- version of Fixture Monkey: 1.1.5
- version of Java: 17

## Steps to reproduce
1. Create a test using FixtureMonkey where BigDecimal or floating-point (float, double) values are generated
2. Set minimum and maximum values to be equal (e.g., 2.0)
3. The borders are not included in the range

```java
class TEST {
    private Validator validator;

    @BeforeEach
    void setUp() {
        ValidatorFactory factory = Validation.buildDefaultValidatorFactory();
        validator = factory.getValidator();
    }

    @Test
    void success() {
        A object = new A(new BigDecimal("2"));

        Set<ConstraintViolation<A>> violations = validator.validate(object);
        assertThat(violations).isEmpty(); // [SUCCESS]
    }

    @Test
    void fail() {
            .plugin(new JakartaValidationPlugin())
            .build();

        sut.giveMeOne(A.class); // [ERROR] If min value [2.0] is equal to max value [2.0] borders must be included.
    }

    public static class A {
        public A() {}

        public A(BigDecimal value) {
            this.value = value;
        }

        @NotNull
        // The same exception occurs with:
        // - @Min @Max annotations
        // - float/double types with similar constraints        
        @DecimalMax(value = "2.0")
        @DecimalMin(value = "2.0")
        private BigDecimal value;     

        public BigDecimal getValue() {
            return value;
        }

        public void setValue( BigDecimal value) {
            this.value = value;
        }
    }
}
```

```
If min value [2.0] is equal to max value [2.0] borders must be included.
java.lang.IllegalArgumentException: If min value [2.0] is equal to max value [2.0] borders must be included.
	at net.jqwik.engine.properties.Range.of(Range.java:18)
	at net.jqwik.engine.properties.arbitraries.DefaultBigDecimalArbitrary.between(DefaultBigDecimalArbitrary.java:57)
	at net.jqwik.engine.properties.arbitraries.DefaultBigDecimalArbitrary.lessOrEqual(DefaultBigDecimalArbitrary.java:68)
	at com.navercorp.fixturemonkey.api.jqwik.JqwikJavaArbitraryResolver.bigDecimals(JqwikJavaArbitraryResolver.java:476)
```

## Expected behaviour
When minimum and maximum values are equal (e.g., 2.0), the test data should be generated successfully.
The generated test data should match the actual validation behavior
This will ensure consistent behavior between test data generation and actual validation logic.

## Actual behaviour
The system throws an IllegalArgumentException during test data generation, preventing the creation of test cases with equal minimum and maximum values. This causes a mismatch between test data generation and actual validation behavior.

**If this exception is not intended behavior, please let me know if I can submit a PR for the fix!**

The root cause appears to be the incorrect handling of the inclusive property internally.

Although I can identify the likely source of the issue, I would appreciate any tips for running integration tests after the fix. In particular, I'd like to make sure the fix doesn't cause any unexpected behavior in other scenarios.

Seol-JY:
## Summary
This PR addresses several issues and enhances the decimal validation handling in Fixture Monkey, primarily focused on fixing #1126 and improving the overall validation behavior for numeric types.

Related issues:
Fixes #1126, #1127

## Description
This PR implements comprehensive improvements to the decimal validation system:

#### 1. Fixed `Inclusive` Property Handling
- Resolved issues with incorrect handling of the `inclusive` property in validation annotations
- Ensures proper behavior when minimum and maximum values are equal
- Applies to `BigDecimal`, `float`, and `double` types

#### 2. Enhanced Multiple Validation Handling
- Implemented proper AND operation logic for multiple constraints
 (e.g., when `@DecimalMax("11.5") @Max(10)` is applied to a field, values should be less than or equal to 10.0, as it satisfies both constraints)
- Ensures all validation conditions are satisfied simultaneously

#### 3. Improved `@Digits` Annotation Processing
- Aligned `@Digits` validation behavior with Jakarta Validation's actual implementation
- Previous implementation was overly restrictive:
  - For `@Digits(integer=3, fraction=2)`, strictly generated numbers in `xxx.xx` format
  - This forced exact digit counts for both integer and fraction parts
- New implementation matches actual validation rules:
  - For `@Digits(integer=3, fraction=2)`, correctly allows:
    - Integer part: any number up to 999 (≤ 3 digits)
    - Fraction part: up to 2 decimal places (not requiring exactly 2)
  - Example: Both 5.4 and 999.99 are now valid where previously only `xxx.xx` format was generated

#### 4. ~~Added Repeatable Annotation Support~~
- ~~Implemented collection of all instances from container annotations marked with `@Repeatable`~~
- ~~Ensures comprehensive validation when multiple constraints are present~~

#### 5. Simplified `JavaDecimalConstraint` Structure

- Reduced redundant fields by removing separate `positive`/`negative` range handling
- New simplified constructor with single `min`/`max` range:
  ```java
  JavaDecimalConstraint(
      @Nullable BigDecimal min,
      @Nullable Boolean minInclusive,
      @Nullable BigDecimal max,
      @Nullable Boolean maxInclusive,
      @Nullable Integer scale
  )
  ```
- Ensures uniform distribution across valid ranges

This change aligns with the `@Digits` validation improvement from No. 3, ensuring proper handling of decimal constraints while simplifying the codebase.

#### 6. `KotestJavaArbitraryGeneratorSet` Improvements
- Aligned with the simplified `JavaDecimalConstraint` structure
- Simplified `inclusive`/`exclusive` boundary handling by leveraging Kotest's default behavior for decimal value generators (`Arb.double()`, `Arb.float()`, `Arb.bigDecimal()`)
  - Instead of waiting for `Kotest` updates to handle inclusive flags separately, implemented a clean solution using `MIN_VALUE` adjustments

#### 7. Code Refactoring
- Improved code organization and readability
- Enhanced error handling and validation logic
- Simplified constraint processing workflow

#### Double, Float, and BigDecimal Validation Tests
- Added tests for equal min/max bounds using `@DecimalMin("10.0")`/`@DecimalMax("10.0")` to verify exact value generation
- Verified handling of multiple constraints (`@DecimalMax("11.5")` with `@Max(10)`) to ensure the stricter bound is applied
- Tested exclusive boundary handling by verifying generated values fall strictly between 10.0 and 10.1 using `inclusive = false`

## Is the Document updated?
No documentation update is required as this is a bug fix for internal implementation.


fixture-monkey-javax-validation/src/test/java/com/navercorp/fixturemonkey/javax/validation/spec/BigDecimalIntrospectorSpec.java
```diff
@@ -52,4 +52,12 @@ public class BigDecimalIntrospectorSpec {
...
 	@PositiveOrZero
 	private BigDecimal positiveOrZero;
+
+	@DecimalMax(value = "100.1")
+	@DecimalMin(value = "100.1")
+	private BigDecimal decimalEqual;
+
+	@Max(value = 100)
+	@Min(value = 100)
+	private BigDecimal integerEqual;
```
seongahjo, line 62: I cannot infer your intent, could you create a new DTO and make a new concise and clear test to reveal your intent?

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/generator/ArbitraryGeneratorContext.java
```diff
@@ -107,6 +111,30 @@ public <T extends Annotation> Optional<T> findAnnotation(Class<T> annotationClas
...
+						Method valueMethod = container.annotationType().getDeclaredMethod("value");
+						T[] values = (T[])valueMethod.invoke(container);
+						results.addAll(Arrays.asList(values));
+					} catch (Exception ignored) {
+					}
+				});
+		}
+
+		return results;
+	}
```
seongahjo, line 136: It is not for common case.

You can not confirm that user uses the annotation `@Repeatable` , the name of property `value` .

I think it should be removed.

Seol-JY: @seongahjo 
Ah, I understand! Instead of defining a new `findAnnotations` in `ArbitraryGeneratorContext`, should I handle duplicate annotations within `ArbitraryGeneratorContext`?

seongahjo: Does `this.getResolvedProperty().getAnnotations()` not return the duplicate annotations?

Could you make a test to show when the duplicate annotation is an issue?

Seol-JY:
```java
@Getter
@Setter
public class TestSpec {
	@Max(value = 50)        // interface jakarta.validation.constraints.Max
	private BigDecimal a;

	@Max(value = 100)       // multiple @Max annotations - internally processed as container annotation
	@Max(value = 50)
	private BigDecimal b;     

	@Max.List({             // interface jakarta.validation.constraints.Max$List
		@Max(value = 100),
		@Max(value = 50)
	})
	private BigDecimal c;
}
```
Duplicate annotations appear to be automatically processed in the same way as the annotations in field c.

Therefore, I implemented `findAnnotations` to process the contents within `Max$List` in the same way as regular duplicate annotations, as shown above.

Without the `Repeatable` processing in `findAnnotations`, if we only use `this.getResolvedProperty().getAnnotations()`, we would need to search twice using `getAnnotations(Max.class)` and `getAnnotations(Max.List.class)`. This is because we need to consider both cases where there is a single annotation and multiple annotations.
Since `getAnnotationsByType` cannot be used in the current implementation, I chose to implement `findAnnotations` to handle both single annotations and container annotations in a unified way.

seongahjo: Does the JSR-303 support multiple annotations on the same property within the same group?

I only found the applying multiple constraint within the different group.
https://beanvalidation.org/1.0/spec/#constraintsdefinitionimplementation-multipleconstraints

Seol-JY: Since the @Repeatable annotation allows multiple annotations to be applied, it was confirmed during testing that all validation annotations (including duplicates) are processed as expected.

However, JSR-303 does not define a specific approach for handling duplicate annotations, meaning implementations can vary.

Given this, rather than implementing complex logic to handle duplicates, it may be more practical to revert to the original approach of simply checking for the presence of annotations.

fixture-monkey-jakarta-validation/src/main/java/com/navercorp/fixturemonkey/jakarta/validation/introspector/JakartaValidationConstraintGenerator.java
```diff
@@ -292,178 +292,126 @@ public JavaIntegerConstraint generateIntegerConstraint(ArbitraryGeneratorContext
...
+			isPositiveMin ? minInclusive : null,
+			isPositiveMax ? max : null,
+			isPositiveMax ? maxInclusive : null,
+
+			isNegativeMin ? min : null,
+			isNegativeMin ? minInclusive : null,
+			isNegativeMax ? max : null,
+			isNegativeMax ? maxInclusive : null,
 			scale
 		);
```
Seol-JY, line 416: I don't understand why the JavaDecimalConstraint class needs to manage separate min/max values for negative and positive numbers. In the previous implementation with the @Digits annotation, this was necessary, but with the "3. Improved @Digits Annotation Processing," the method for enforcing maximum digit limits has changed. In this case, wouldn't it be sufficient to just manage the min/max values and their inclusiveness to generate random numbers? What do you think about changing this approach?

seongahjo: Is there a better idea for addressing both positive and negative values in the case of `@Digits`?

For example, in the case of `@Digits(integer=3, fraction=2)`, it should be the value that satisfies the following constraints.

1. 0 <= value <= 999.99
2. -999.99 <= value <= 0

Seol-JY: Looking at your example with `@Digits(integer=3, fraction=2)`, I believe we could satisfy these constraints simply using a minimum (-999.99) and maximum (999.99) value.
Is there any specific case where we need to manage positive and negative ranges separately in the current implementation? If I'm missing something important, please let me know.

Seol-JY: To maintain compatibility with the existing structure (as it seems there would be many affected areas due to changes), I decided to continue using the existing `JavaDecimalConstraint`.

seongahjo: Sorry for the delay. I have misunderstood the validation with `@Digits`.
It is ok to set the limit within a minimum (-999.99) and maximum (999.99) value.

> To maintain compatibility with the existing structure (as it seems there would be many affected areas due to changes)

Can you tell me what the breaking change is?

fixture-monkey-jakarta-validation/src/main/java/com/navercorp/fixturemonkey/jakarta/validation/introspector/JakartaValidationConstraintGenerator.java
```diff
@@ -292,180 +292,158 @@ public JavaIntegerConstraint generateIntegerConstraint(ArbitraryGeneratorContext
...
 
-		if (positiveMin == null && positiveMax == null && negativeMin == null && negativeMax == null && scale == null) {
+		if (min == null && max == null) {
 			return null;
 		}
 
+		boolean isPositiveMin = min != null && min.compareTo(BigDecimal.ZERO) >= 0;
+		boolean isPositiveMax = max != null && max.compareTo(BigDecimal.ZERO) >= 0;
+		boolean isNegativeMin = min != null && min.compareTo(BigDecimal.ZERO) < 0;
+		boolean isNegativeMax = max != null && max.compareTo(BigDecimal.ZERO) < 0;
```
seongahjo, line 405: NIT, in convention of FixtureMonkey, only the boolean parameter allows the prefix of `is`.

fixture-monkey-jakarta-validation/src/main/java/com/navercorp/fixturemonkey/jakarta/validation/introspector/JakartaValidationConstraintGenerator.java
Seol-JY: No problem! The `JavaDecimalConstraint` class appears to be used in the following methods:
1. ValidateArbitraryGenerator#generate()
2. JqwikJavaArbitraryResolver#floats()
3. JqwikJavaArbitraryResolver#doubles()
4. JqwikJavaArbitraryResolver#bigDecimals()
5. SimpleValueJqwikPlugin.SimpleJavaConstraintGenerator#generateDecimalConstraint()

I want to reduce the fields in `JavaDecimalConstraint` as follows:
```java
public JavaDecimalConstraint(
    @Nullable BigDecimal min
    @Nullable Boolean minInclusive,
    @Nullable BigDecimal max,
    @Nullable Boolean maxInclusive,
    @Nullable Integer scale
) {
    this.min = min;
    this.minInclusive = minInclusive;
    this.max = max;
    this.maxInclusive = maxInclusive;
    this.scale = scale;
}
```
Would it be okay to make this change and modify the implementation of related methods? I'm a bit concerned about potential side effects. Also, I'm curious whether JavaDecimalConstraint is also being used in any Kotlin implementations!

seongahjo: I think it might not be a problem. It is actually used in Kotlin, but it does not produce the side effect.

The tests will fail if breaking change happens.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/plugin/SimpleValueJqwikPlugin.java
```diff
@@ -278,13 +278,9 @@ public JavaDecimalConstraint generateDecimalConstraint(ArbitraryGeneratorContext
 			}
 
 			return new JavaDecimalConstraint(
-				positiveMin,
+				BigDecimal.valueOf(this.negativeMinNumberValue),
```
seongahjo, line 281: `SimpleValueJqwikPlugin` needs to remove the separation between positive and negative.

It will not work if the min value is set to 1 and the max value is set to 10.

seongahjo: > It will not work if the min value is set to 1 and the max value is set to 10.

It would be helpful to add the above test case.

Seol-JY: I created test cases for decimal number ranges. Please let me know if we need tests for other decimal types or if the current test cases seem excessive!

seongahjo (approved):
Thank you for your effort!
The contribution is self-explanatory and well tested.

PR #1146, opened 2025-01-18
Add support for deterministic testing with JUnit

seongahjo:
## Summary
Add support for deterministic testing with JUnit

## (Optional): Description
*Describe your changes in detail*

## How Has This Been Tested?
* existing tests

## Is the Document updated?
Later


fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/engine/EngineUtils.java
```diff
@@ -0,0 +1,42 @@
+package com.navercorp.fixturemonkey.api.engine;
```
acktsap, line 1: Seems no license comment

PR #1179, opened 2025-04-22
Add StringCombinableArbitrary for easy efficient customisation

seongahjo:
## Summary
Add StringCombinableArbitrary for easy efficient customisation

## (Optional): Description
providing below APIs to customize the randomly generated String

- alphabetic
- korean
- ascii
- numeric
- pattern
- withLength
- withMinLength
- withMaxLength

## How Has This Been Tested?
* stringCombinableArbitraryIsJqwik
* stringCombinableArbitraryInjectNull
* stringCombinableArbitraryFilter
* stringCombinableArbitraryFilterCharacter
* stringCombinableArbitraryFilterCharacterAndFilter
* stringCombinableArbitraryNumeric
* stringCombinableArbitraryMap
* stringCombinableArbitraryKorean
* stringCombinableArbitraryAlphabet
* stringCombinableArbitraryLatterWins
* stringCombinableArbitrary
* stringCombinableArbitraryAscii

## Is the Document updated?
Later

### DIFF
--- a/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/StringCombinableArbitrary.java
+++ b/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/StringCombinableArbitrary.java
@@ -1,0 +1,119 @@
+/*
+ * Fixture Monkey
+ *
+ * Copyright (c) 2021-present NAVER Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.navercorp.fixturemonkey.api.arbitrary;
+
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+
+public interface StringCombinableArbitrary extends CombinableArbitrary<String> {
+	int STRING_DEFAULT_MIN_LENGTH = 0;
+	int STRING_DEFAULT_MAX_LENGTH = 100;
+
+	@Override
+	String combined();
+
+	@Override
+	String rawValue();
+
+	StringCombinableArbitrary withLength(int min, int max);
+
+	/**
+	 * Generates a StringCombinableArbitrary which contains only alphabetic characters.
+	 * It conflicts with {@link #ascii()} and {@link #numeric()} and {@link #korean()}.
+	 *
+	 * @return the StringCombinableArbitrary which contains only alphabetic characters
+	 */
+	StringCombinableArbitrary alphabetic();
+
+	/**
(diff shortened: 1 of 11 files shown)


fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/StringCombinableArbitrary.java
```diff
@@ -0,0 +1,119 @@
...
+
+	/**
+	 * Generates the pattern matched arbitrary String.
+	 * Its performance differs from the generation engine you're using.
+	 *
+	 * @param stringPattern the regular expression pattern
+	 * @return the pattern matched StringCombinableArbitrary
+	 */
+	default StringCombinableArbitrary pattern(int tries, String stringPattern) {
+		Pattern pattern = Pattern.compile(stringPattern);
```
acktsap, line 100: How about adding method taking `Pattern` class? It seems [immutable](https://github.com/openjdk/jdk/blob/c873837da6e373613866f5f5c0017f0fccb97b57/src/java.base/share/classes/java/util/regex/Pattern.java#L77)

fixture-monkey-api/src/main/resources/META-INF/services/com.navercorp.fixturemonkey.api.arbitrary.StringCombinableArbitrary
```diff
@@ -0,0 +1 @@
+com.navercorp.fixturemonkey.api.jqwik.JqwikStringCombinableArbitrary
```
acktsap, line 1: Seems it's already loaded (it's inside the api module). Is that ok?

fixture-monkey-kotest/src/main/kotlin/com/navercorp/fixturemonkey/kotest/KotestStringCombinableArbitrary.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    override fun filter(tries: Int, predicate: Predicate<String>): StringCombinableArbitrary =
+        KotestStringCombinableArbitrary(arb.filter { predicate.test(it) })
+
+    override fun withLength(min: Int, max: Int): StringCombinableArbitrary =
+        KotestStringCombinableArbitrary(Arb.string(min..max))
+
+    override fun alphabetic(): StringCombinableArbitrary = KotestStringCombinableArbitrary(
+        Arb.string(
+            codepoints = Arb.codepoints().filter { it.value.toChar() in 'a'..'z' }
```
acktsap, line 43: Is it ok to include lower case only?

fixture-monkey-api/src/main/resources/META-INF/services/com.navercorp.fixturemonkey.api.arbitrary.StringCombinableArbitrary
seongahjo: As I test it, SPI does not load the implementation implicitly.
It should specify the implementation in `META-INF/services`.

I also did not find the following specification in the document.

https://docs.oracle.com/javase/tutorial/sound/SPI-intro.html

image

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/StringCombinableArbitrary.java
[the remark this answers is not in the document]
seongahjo: The 'filter' API may conflict with other APIs by design. Therefore it is better to add a document explaining it.

Issue #1188, opened 2025-05-28
Add IntegerCombinableArbitrary for easy Integer customization


seongahjo:
## Summary
Add `IntegerCombinableArbitrary` to allow easy customization of randomly generated Integer values, similar to the existing `StringCombinableArbitrary`.

## Description
Currently, when generating Integer values with Fixture Monkey, users cannot easily customize the generation behavior (like generating only positive numbers, or numbers within a specific range).

We recently added `StringCombinableArbitrary` in PR #1179 which allows easy string customization with methods like `.alphabetic()`, `.numeric()`, `.withLength()`, etc.

We need similar functionality for Integer type with methods like:
- `.positive()` - only positive numbers
- `.negative()` - only negative numbers  
- `.withRange(min, max)` - numbers within range
- `.even()` - only even numbers
- `.odd()` - only odd numbers

## Expected Usage
```java
// Current way (limited customization)
Integer number = fixtureMonkey.giveMeOne(Integer.class);

// Proposed way (easy customization)
Integer positiveNumber = CombinableArbitrary.integers().positive().combined();
Integer evenNumber = CombinableArbitrary.integers().even().combined();
Integer rangedNumber = CombinableArbitrary.integers().withRange(1, 100).combined();
```

## Implementation Hints
1. Look at how `StringCombinableArbitrary` is implemented in PR #1179
2. Create `IntegerCombinableArbitrary` interface following the same pattern
3. Create Jqwik implementation (`JqwikIntegerCombinableArbitrary`)
4. Add factory method to `CombinableArbitrary.integers()`
5. Add tests similar to the string tests in `JavaTest.java`

## Files to Reference
- `StringCombinableArbitrary.java` - main interface pattern
- `JqwikStringCombinableArbitrary.java` - implementation example
- `CombinableArbitrary.java` - where to add the factory method
- `JavaTest.java` - test examples

## Good First Issue Because
- Clear pattern to follow from existing `StringCombinableArbitrary`
- Well-defined scope (just Integer type)
- Good introduction to the codebase architecture
- Plenty of existing examples to reference

## If you're interested
1. Comment on this issue to get assigned
2. Look at PR #1179 to understand the pattern
3. Feel free to ask questions in the comments!

PR #1192, opened 2025-05-31
Add IntegerCombinableArbitrary for easy Integer customization

Denia-park:
## Summary
Add `IntegerCombinableArbitrary` to allow easy customization of randomly generated Integer values, similar to the existing `StringCombinableArbitrary`. (#1188)

## (Optional): Description
providing below APIs to customize the randomly generated Integer

`.positive()` - only positive numbers
`.negative()` - only negative numbers
`.withRange(min, max)` - numbers within range
`.even()` - only even numbers
`.odd()` - only odd numbers

## How Has This Been Tested?
- `integerCombinableArbitraryIsJqwik`
- `integerCombinableArbitraryInjectNull`
- `integerCombinableArbitraryFilter`
- `integerCombinableArbitraryPositive`
- `integerCombinableArbitraryNegative`
- `integerCombinableArbitraryWithRange`
- `integerCombinableArbitraryWithRangeAndFilter`
- `integerCombinableArbitraryEven`
- `integerCombinableArbitraryOdd`
- `integerCombinableArbitraryLastOperationWinsWithPositiveAndNegative`
- `integerCombinableArbitraryLastOperationWinsWithEvenAndOdd`
- `integerCombinableArbitraryLastOperationWinsWithNegativeAndRange`

## Is the Document updated?
No. Once all code reviews are complete, would it be OK if I update and push the documentation just before merging this PR? 
If that is not acceptable, I will update the documentation right away. Thank you for your understanding.

### DIFF
--- a/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/IntegerCombinableArbitrary.java
+++ b/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/IntegerCombinableArbitrary.java
@@ -1,0 +1,92 @@
+/*
+ * Fixture Monkey
+ *
+ * Copyright (c) 2021-present NAVER Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.navercorp.fixturemonkey.api.arbitrary;
+
+import java.util.function.Predicate;
+
+public interface IntegerCombinableArbitrary extends CombinableArbitrary<Integer> {
+	@Override
+	Integer combined();
+
+	@Override
+	Integer rawValue();
+
+	/**
+	 * Generates an IntegerCombinableArbitrary which produces integers within the specified range.
+	 *
+	 * @param min the minimum value (inclusive)
+	 * @param max the maximum value (inclusive)
+	 * @return the IntegerCombinableArbitrary producing integers between {@code min} and {@code max}
+	 */
+	IntegerCombinableArbitrary withRange(int min, int max);
+
+	/**
+	 * Generates an IntegerCombinableArbitrary which produces only positive integers.
+	 *
+	 * @return the IntegerCombinableArbitrary producing positive integers
+	 */
+	IntegerCombinableArbitrary positive();
+
+	/**
+	 * Generates an IntegerCombinableArbitrary which produces only negative integers.
+	 *
+	 * @return the IntegerCombinableArbitrary producing negative integers
+	 */
+	IntegerCombinableArbitrary negative();
+
+	/**
+	 * Generates an IntegerCombinableArbitrary which produces only even integers.
+	 *
+	 * @return the IntegerCombinableArbitrary producing even integers
+	 */
+	IntegerCombinableArbitrary even();
+
+	/**
+	 * Generates an IntegerCombinableArbitrary which produces only odd integers.
+	 *
+	 * @return the IntegerCombinableArbitrary producing odd integers
+	 */
+	IntegerCombinableArbitrary odd();
+
+	@Override
+	default IntegerCombinableArbitrary filter(Predicate<Integer> predicate) {
+		return this.filter(DEFAULT_MAX_TRIES, predicate);
+	}
+
+	@Override
+	default IntegerCombinableArbitrary filter(int tries, Predicate<Integer> predicate) {
+		return new IntegerCombinableArbitraryDelegator(CombinableArbitrary.super.filter(tries, predicate));
+	}
+
+	@Override
+	default IntegerCombinableArbitrary injectNull(double nullProbability) {
+		return new IntegerCombinableArbitraryDelegator(CombinableArbitrary.super.injectNull(nullProbability));
+	}
+
+	@Override
+	default IntegerCombinableArbitrary unique() {
+		return new IntegerCombinableArbitraryDelegator(CombinableArbitrary.super.unique());
+	}
+
+	@Override
+	void clear();
+
+	@Override
+	boolean fixed();
+}
--- a/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/IntegerCombinableArbitrary.java
+++ b/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/IntegerCombinableArbitrary.java
@@ -29,7 +29,7 @@ public interface IntegerCombinableArbitrary extends CombinableArbitrary<Integer>
 	Integer combined();
 
 	@Override
-	IntegerCombinableArbitrary rawValue();
+	Integer rawValue();
 
 	/**
 	 * Generates an IntegerCombinableArbitrary which produces integers within the specified range.
--- a/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/IntegerCombinableArbitrary.java
+++ b/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/IntegerCombinableArbitrary.java
@@ -20,12 +20,16 @@ package com.navercorp.fixturemonkey.api.arbitrary;
 
 import java.util.function.Predicate;
 
+import org.apiguardian.api.API;
+import org.apiguardian.api.API.Status;
+
+@API(since = "1.1.12", status = Status.EXPERIMENTAL)
 public interface IntegerCombinableArbitrary extends CombinableArbitrary<Integer> {
 	@Override
 	Integer combined();
 
 	@Override
-	Integer rawValue();
+	IntegerCombinableArbitrary rawValue();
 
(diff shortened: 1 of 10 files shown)


fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/IntegerCombinableArbitrary.java
```diff
@@ -0,0 +1,92 @@
...
+package com.navercorp.fixturemonkey.api.arbitrary;
+
+import java.util.function.Predicate;
+
+public interface IntegerCombinableArbitrary extends CombinableArbitrary<Integer> {
+	@Override
+	Integer combined();
+
+	@Override
+	Integer rawValue();
```
seongahjo, line 32: It is better to only override the return type to be set to 'IntegerCombinableArbitrary'.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/IntegerCombinableArbitrary.java
```diff
@@ -0,0 +1,92 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.navercorp.fixturemonkey.api.arbitrary;
+
+import java.util.function.Predicate;
+
+public interface IntegerCombinableArbitrary extends CombinableArbitrary<Integer> {
```
seongahjo, line 27: It is better to annotate with `@API(since = "1.1.12", status = Status.EXPERIMENTAL)`

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/jqwik/JqwikIntegerCombinableArbitrary.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+package com.navercorp.fixturemonkey.api.jqwik;
+
+import net.jqwik.api.Arbitraries;
+import net.jqwik.api.Arbitrary;
+
+import com.navercorp.fixturemonkey.api.arbitrary.IntegerCombinableArbitrary;
+
+public final class JqwikIntegerCombinableArbitrary implements IntegerCombinableArbitrary {
```
seongahjo, line 30: It is better to annotate with @API(since = "1.1.12", status = Status.EXPERIMENTAL)

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/jqwik/JqwikIntegerCombinableArbitrary.java
```diff
@@ -0,0 +1,83 @@
...
+
+	@Override
+	public Integer rawValue() {
+		return this.combined();
+	}
+
+	@Override
+	public IntegerCombinableArbitrary withRange(int minValue, int maxValue) {
+		return new JqwikIntegerCombinableArbitrary(
+			Arbitraries.integers().filter(it -> minValue <= it && it <= maxValue)
```
seongahjo: For better performance, it is better to use Arbitraries.integers().between().

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/jqwik/JqwikIntegerCombinableArbitrary.java
```diff
@@ -0,0 +1,83 @@
...
+		);
+	}
+
+	@Override
+	public IntegerCombinableArbitrary positive() {
+		return new JqwikIntegerCombinableArbitrary(Arbitraries.integers().filter(it -> 0 < it));
```
seongahjo: For better performance, it is better to use 
Arbitraries.integers().greaterOrEqual().

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/jqwik/JqwikIntegerCombinableArbitrary.java
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	@Override
+	public IntegerCombinableArbitrary negative() {
+		return new JqwikIntegerCombinableArbitrary(Arbitraries.integers().filter(it -> it < 0));
```
seongahjo: For better performance, it is better to use 
Arbitraries.integers().lessOrEqual().

fixture-monkey-tests/java-tests/src/test/java/com/navercorp/fixturemonkey/tests/java/JavaTest.java
```diff
@@ -1524,4 +1525,88 @@ void registerSizeLessThanThree() {
...
 	}
+
+	@Test
+	void integerCombinableArbitraryIsJqwik() {
+		CombinableArbitrary<Integer> actual = CombinableArbitrary.integers();
+
+		then(actual).isInstanceOf(JqwikIntegerCombinableArbitrary.class);
+	}
+
+	@Test
```
seongahjo: It is better to test repeatedly.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/arbitrary/IntegerCombinableArbitrary.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apiguardian.api.API;
+import org.apiguardian.api.API.Status;
+
+@API(since = "1.1.12", status = Status.EXPERIMENTAL)
+public interface IntegerCombinableArbitrary extends CombinableArbitrary<Integer> {
+	@Override
+	Integer combined();
+
+	@Override
+	IntegerCombinableArbitrary rawValue();
```
seongahjo: It should return `Object`, you do not have to override `rawValue()`

Denia-park: @seongahjo 

When I don’t override the `rawValue()` method in the `IntegerCombinableArbitrary` interface, 
could you please clarify how the following classes are expected to implement `rawValue()` and what they should return?
- `IntegerCombinableArbitraryDelegator`
- `JqwikIntegerCombinableArbitrary`
- `KotestIntegerCombinableArbitrary`

---

I’m not entirely sure what value `rawValue()` is supposed to provide, so I’m having trouble understanding how to implement it correctly. 
Could you also explain if there’s something I’m missing?

seongahjo: @Denia-park 

I was mistaken, it should override as well. All the implementations should return an `Integer`.

The method rawValue() returns the actual, uncustomised instance of the type, ensuring compatibility with deserialization.

Jackson, for example, is the right case.

As each CombinableArbitrary is responsible for generating an arbitrary value, each one should deserialize the JSON.

In that case, We cannot support the jackson specialized annotations, `@JsonTypeInfo`, `@JsonSubTypes`.

The `JacksonCombinableArbitrary` class and the tests using `JacksonSpecs` would be helpful.

Denia-park: @seongahjo 

i update my code.

denia-hgpark:
Hello @Seol-JY

Thank you very much for your detailed review on PR.

Below, I answer each question you raised in the comments.

## 1. “Should Last Wins apply to compatible conditions?”
``` java
@Override
public IntegerCombinableArbitrary positive() {
    return new JqwikIntegerCombinableArbitrary(
        Arbitraries.integers()..greaterOrEqual(1)
    );
}
```
All specialized methods — `positive()`, `negative()`, `withRange()`, `even()`, `odd()` — always start from a new `Arbitraries.integers()` stream. 

Therefore:
```java
withRange(-10, 10).positive()   // Result: (0‥10)
positive().withRange(-10, 10)   // Result: (-10‥10)  ← Last call takes precedence
```
In other words, the rule “the most recently called method overrides all previous conditions (Last Wins)” naturally holds.

> **Reason for adding the test case**
> Because each specialized method of `IntegerCombinableArbitrary` always creates a new stream, i added this test to clearly demonstrate that the operators invoked later take precedence.

## 2. “Inconsistent behavior with filters”
Call Order | Works? | Explanation
-- | -- | --
`.positive().filter(i -> i < 100)` | ✅ Yes | `filter` is applied last, layering an extra constraint on top of the existing `positive()`.
`.filter(i -> i > 0).positive()` | ❌ No | `positive()` returns a new instance → the earlier filter condition is lost.

As shown above, filter accumulates constraints on the existing instance, whereas the specialized methods (`positive()`, `negative()` ...) return a new instance, causing the behavior to vary depending on call order.

## 3. Questions for Discussion
@seongahjo 
i believe this part requires further discussion. 

How would you like it to be structured? 

Personally, I think they should be usable in combination. (For example: `.positive().odd() // Possible combination (1,3,5,7...)` )

---

I will apply any feedback promptly. Thank you!

Denia-park:
@YongGoose 

> If we want to apply all selected operations instead of relying on a `last wins` strategy, it might be possible to store the `integerArbitrary` instances in order within `JqwikIntegerCombinableArbitrary`.
> 
> However, I suspect there could be conflicts in cases where conflicting operations like `.positive()` and `.negative()` are applied together.

You’re absolutely right. 

At first, I tried to implement them so they would all work together, 
but I got confused by directly opposing methods such as `odd()` ↔ `even()` or `positive()` ↔ `negative()`.

So for now, I’ve decided to implement each method as a standalone feature. 😂

PR #1201, opened 2025-06-22
Fix BeanArbitraryIntrospector - propertyDescriptor getWriteMethod throw NPE 

bbeomgeun:
## Summary

To use BeanArbitraryIntrospector, both a default constructor and setter methods are currently required.
For the case of a missing default constructor, a warning log is printed, but no exception is thrown.
However, when a setter method is missing, a NullPointerException occurs.

Related issues:
Fixes https://github.com/naver/fixture-monkey/issues/1199

## (Optional): Description

The NPE is thrown at the point where propertyDescriptor.getWriteMethod() is invoked
- this happens because the corresponding setXXX method does not exist.

Although a null check on propertyDescriptor can be added before calling getWriteMethod(), it may not be sufficient since propertyDescriptor can still be non-null if a getter exists, even when the setter is missing.
Therefore, I modified the logic to catch the exception after invoking the setter method and added a warning log instead.

## How Has This Been Tested?

objectGenerationShouldWorkWithoutSetter
- Add NoSetterSpecs

**To cover the case where a setter is missing, I’ve created a Specs.
Could you please review it?**

## Is the Document updated?

N/A

### DIFF
--- a/fixture-monkey-tests/java-tests/src/test/java/com/navercorp/fixturemonkey/tests/java/BeanArbitraryIntrospectorTest.java
+++ b/fixture-monkey-tests/java-tests/src/test/java/com/navercorp/fixturemonkey/tests/java/BeanArbitraryIntrospectorTest.java
@@ -51,7 +51,7 @@ class BeanArbitraryIntrospectorTest {
 
 	@RepeatedTest(TEST_COUNT)
 	void objectGenerationShouldWorkWithoutSetter() {
-		NoSetterSpecs.JavaTypeObject actual = SUT.giveMeOne(NoSetterSpecs.JavaTypeObject.class);
+		NoSetterSpecs.StringObject actual = SUT.giveMeOne(NoSetterSpecs.StringObject.class);
 
 		then(actual).isNotNull();
 	}
--- a/fixture-monkey-tests/java-tests/src/test/java/com/navercorp/fixturemonkey/tests/java/specs/NoSetterSpecs.java
+++ b/fixture-monkey-tests/java-tests/src/test/java/com/navercorp/fixturemonkey/tests/java/specs/NoSetterSpecs.java
@@ -6,36 +6,7 @@ import lombok.NoArgsConstructor;
 public class NoSetterSpecs {
 	@NoArgsConstructor
 	@Getter
-	public static class JavaTypeObject {
-		private String string;
-		private int primitiveInteger;
-		private float primitiveFloat;
-		private long primitiveLong;
-		private double primitiveDouble;
-		private byte primitiveByte;
-		private char primitiveCharacter;
-		private short primitiveShort;
-		private boolean primitiveBoolean;
-		private Integer wrapperInteger;
-		private Float wrapperFloat;
-		private Long wrapperLong;
-		private Double wrapperDouble;
-		private Byte wrapperByte;
-		private Character wrapperCharacter;
-		private Short wrapperShort;
-		private Boolean wrapperBoolean;
-		private Enum enumValue;
-	}
-
-	@NoArgsConstructor
-	@Getter
 	public static class StringObject {
 		private String value;
-	}
-
-	@NoArgsConstructor
-	@Getter
-	public static class NestedObject {
-		private StringObject object;
 	}
 }


fixture-monkey-tests/java-tests/src/test/java/com/navercorp/fixturemonkey/tests/java/BeanArbitraryIntrospectorTest.java
```diff
@@ -47,4 +48,11 @@ void setPropertyPostCondition() {
 
 		then(actual).hasSizeGreaterThan(5);
 	}
+
+	@RepeatedTest(TEST_COUNT)
+	void objectGenerationShouldWorkWithoutSetter() {
```
seongahjo, line 53: It is much better to write a test case that describes the simplest possible version of the issue.

In this case, it would be `StringObject`.

The `JavaTypeObject` and `NestedObject` are not needed.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/introspector/BeanArbitraryIntrospector.java
```diff
@@ -117,6 +117,10 @@ private Function<Map<ArbitraryProperty, Object>, Object> combine(
 							writeMethod.getName(),
 							value,
 							ex);
+					} catch (NullPointerException ex) {
+						LOGGER.warn("Invoke writeMethod is failed. The given property '{}' has no setter",
```
seongahjo: I think the user shouldn't need to know about the `writeMethod`.

How about adding a notice to check whether the `property` (variable, actually is class) has a setter?

bbeomgeun: Good. I agree that it's enough to just check whether the property has a setter. I'll fix it

fixture-monkey-tests/java-tests/src/test/java/com/navercorp/fixturemonkey/tests/java/BeanArbitraryIntrospectorTest.java
bbeomgeun: Okay. How about adding a `NoSetterSpecs` class?
Personally, I think that might be too much for this case. It might be cleaner to just add a simple inner class within the test case. 
Or just keep StringObject in `NoSetterSpecs` and remove the rest (`JavaTypeObject` and `NestedObject`)
what do you think?

+ I'm trying to figure out the existing test code style as I add to it, so I have quite a few question.

seongahjo: The latter would be better.

It may seem excessive, but it could prevent arbitrary specifications from being made without classification.

bbeomgeun: Good. I removed the rest object in `NoSetterSpecs` at [0d5138a](https://github.com/naver/fixture-monkey/pull/1201/commits/0d5138a58c49afdbb967a80606c31ab951819d4f)

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/introspector/BeanArbitraryIntrospector.java
bbeomgeun: I fixed log message at [fc940bc](https://github.com/naver/fixture-monkey/pull/1201/commits/fc940bc2297f4c45da232dc66499ff93e03ab974)

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/introspector/BeanArbitraryIntrospector.java
```diff
@@ -117,6 +117,10 @@ private Function<Map<ArbitraryProperty, Object>, Object> combine(
 							writeMethod.getName(),
 							value,
 							ex);
+					} catch (NullPointerException ex) {
+						LOGGER.warn("The given property '{}' may not have a setter",
```
seongahjo, line 121: provides more information, `object.getClass()`, would be helpful.

The user can see which classes have no setter.

seongahjo (approved):

PR #1208, opened 2025-06-27
Fix NPE due to nullableContainer control settings in collection elements

Chanwon-Seo:
## Summary
This PR fixes a bug where nullable settings (e.g., nullableElement(true)) override Jakarta/Javax Validation annotations on collection elements.
It ensures that validation constraints like @NotBlank are always respected for elements within collections, preventing the generation of null or invalid values when explicitly disallowed.

Related Issue: [#1207]

## (Optional): Description
Previously, when FixtureMonkey was configured with `nullableContainer(true)` and nullableElement(true) alongside JakartaValidationPlugin (or JavaxValidationPlugin), collection elements with explicit validation annotations (e.g., List<@NotBlank String>) were being generated as null values. This behavior contradicted the intent of the validation annotations and led to unexpected test failures.

This change introduces a fix that prioritizes the validation constraints defined by Jakarta/Javax Validation annotations over the global nullable settings for container elements. Specifically, it ensures that if an element type is annotated with @NotBlank, @NotNull, @NotEmpty, or has a @Size(min=X) constraint that implies non-null/non-empty, it will always be populated with a valid, non-null value, regardless of the nullableElement setting.

This aligns Fixture Monkey's behavior with the expected contract of validation frameworks and provides more reliable data generation for validated models.

## How Has This Been Tested?
The fix has been verified by adding a new property-based test case that explicitly reproduces the original bug and then confirms its resolution.

A new test class ElementAnnotationIntrospectorTest (or an equivalent name within the Jakarta/Javax validation test suites) has been added. This test class includes sampleElementAnnotations() which asserts the following:

Non-empty and Size constraints: then(actual.getSizeContainerAndNotBlankElement()).isNotEmpty(); and then(actual.getSizeContainerAndNotBlankElement()).hasSizeBetween(1, 3);

Element-level @NotBlank constraint: then(actual.getSizeContainerAndNotBlankElement()).allSatisfy(it -> then(it).isNotBlank());

Container @NotEmpty and Element @NotBlank constraints: `then(actual.getNotEmptyContainerAndNotBlankElement()).isNotEmpty();`and `then(actual.getNotEmptyContainerAndNotBlankElement()).allSatisfy(it -> then(it).isNotBlank());`

All existing and new tests pass after applying this change, confirming that the issue is resolved and no regressions were introduced.

## Is the Document updated?
N/A


fixture-monkey-jakarta-validation/src/main/java/com/navercorp/fixturemonkey/jakarta/validation/generator/JakartaValidationNullInjectGenerator.java
```diff
@@ -73,8 +75,45 @@ public double generate(ObjectPropertyGeneratorContext context) {
...
+
+				if (context.getOwnerProperty().isContainer()) {
+					if (elementAnnotations.contains(NotEmpty.class)) {
+						return 0.0d;
+					}
+
+					Size elementSizeAnnotation = elementProperty.getAnnotation(Size.class).orElse(null);
+					if (elementSizeAnnotation != null && elementSizeAnnotation.min() > 0) {
+						return 0.0d;
+					}
```
seongahjo: The following comment suggests that it could be null.
As far as I know, the `@Size` annotation is not related to nullability.

image

seongahjo: You can check the specifications of the annotation below.

image

fixture-monkey-jakarta-validation/src/main/java/com/navercorp/fixturemonkey/jakarta/validation/generator/JakartaValidationNullInjectGenerator.java
```diff
@@ -73,8 +75,45 @@ public double generate(ObjectPropertyGeneratorContext context) {
...
-			return 0.0d;
+		if (context.isContainer()) {
+			if (annotations.contains(NotEmpty.class)) {
+				return 0.0d;
+			}
+
+			Size sizeAnnotation = context.getProperty().getAnnotation(Size.class).orElse(null);
+			if (sizeAnnotation != null && sizeAnnotation.min() > 0) {
+				return 0.0d;
+			}
```
seongahjo: It should be checked https://github.com/naver/fixture-monkey/pull/1208/files#r2174507127

PR #1223, opened 2025-07-25
Fix validate container expression for size in strict mode

### LINKED ISSUE #1219: Missing Validation for `size()` on Container Element Index Expressions
## Describe the bug

`size(notEmptyContainer[0]`와 같은 컨테이너 내부의 요소 인덱스 표현식에 대해 size를 지정할 때 랜덤한 크기(1 ~ 3)로 생성됩니다.

## Your environment

* version of Fixture Monkey: 1.1.4
* version of Java/Kotlin: JDK21

## Steps to reproduce

``` java
FixtureMonkey SUT = FixtureMonkey.builder()
		.plugin(new JakartaValidationPlugin())
		.build();

	@Property
	void logSizeValidationError() {
		thenThrownBy(
			() -> SUT.giveMeBuilder(RootJavaGetterObject.class)
				.size("notEmptyContainer[0]", 1) // [*]
		).isExactlyInstanceOf(IllegalArgumentException.class)
	}

	@Getter
	@Setter
	public static class RootJavaGetterObject {
		private List<@NotEmpty String> notEmptyContainer;
	}
```

## Expected behaviour

`notEmptyContainer[0]`와 같은 컨테이너 요소 인덱스 표현식에 대해 size를 지정할 때 `IllegalArgumentException`이 발생해야 한다고 기대합니다.

## Actual behaviour

예외가 발생하지 않고, `notEmptyContainer`의 크기가 `1 ~ 3` 범위에서 생성됩니다.
이는 표현식에 대해 검증이 이루어지지 않음을 의미하며, 기대하는 동작과는 차이가 있습니다.

해당 문제점이 확인된다면 이 이슈에 대한 PR을 제출할 의향이 있습니다.

Chanwon-Seo:
## Summary
Fix strict mode so that container expression validation (e.g., size) throws exceptions for invalid paths, and add corresponding tests.

Related issues:
Fixes #1219

## (Optional): Description
- Previously, strict mode only validated set/setNull expressions, but did not apply the same validation to size/container expressions.
- This PR ensures that size, minSize, maxSize, etc. also throw exceptions for invalid container expressions in strict mode, making the behavior consistent.
- Added and improved tests to verify that strict mode throws the correct exceptions for invalid container expressions.
- Ensured that exception messages are clear and consistent.

## How Has This Been Tested?
- Added property-based and unit tests to verify that:
  - Invalid container expressions (e.g., size on non-existent fields) in strict mode throw the expected exceptions.
  - set/setNull and size operations now behave consistently in strict mode.
  - Non-strict mode does not throw exceptions for invalid expressions.

## Is the Document updated?
N/A


fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/ExpressionPathValidator.java
```diff
@@ -0,0 +1,48 @@
...
+import java.lang.reflect.Field;
+import java.util.List;
+
+import com.navercorp.fixturemonkey.api.type.Types;
+import com.navercorp.fixturemonkey.tree.NextNodePredicate;
+import com.navercorp.fixturemonkey.tree.PropertyNameNodePredicate;
+
+public class ExpressionPathValidator {
+
+	public static boolean isValidFieldPath(Class<?> rootClass, List<NextNodePredicate> predicates) {
```
seongahjo, line 30: does it work for the `List<List<SimpleObject>>` case?

fixture-monkey/src/test/java/com/navercorp/fixturemonkey/test/FixtureMonkeyOptionsTest.java
```diff
@@ -135,6 +159,16 @@ void notStrictModeSetWrongExpressionDoesNotThrows() {
 				.sample());
 	}
 
+	@Property
+	void notStrictModeSizeWrongExpressionDoesNotThrows() {
```
seongahjo, line 223: Adding the new test case below may be helpful.

```
FixtureMonkey sut = FixtureMonkey.builder().useExpressionStrictMode().build();

val actual = sut.giveMeBuilder(new TypeReference<List<List<SimpleObject>>>(){})
   .size("$", 1)
    .size("$[0]", 1)
    .set("$[0][0].string", "expected")
    .sample().get(0).get(0).getString();

then(actual).isEqualTo("expected);
```

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/ExpressionPathValidator.java
```diff
@@ -0,0 +1,48 @@
...
+		if (predicates == null || predicates.isEmpty()) {
+			return false;
+		}
+		Class<?> currentClass = Types.getActualType(rootClass);
+		for (NextNodePredicate predicate : predicates) {
+			if (predicate instanceof PropertyNameNodePredicate) {
+				String fieldName = ((PropertyNameNodePredicate)predicate).getPropertyName();
+				try {
+					Field field = currentClass.getDeclaredField(fieldName);
+					currentClass = field.getType();
```
seongahjo: What about the property with `@JsonProperty`?

fixture-monkey/src/test/java/com/navercorp/fixturemonkey/test/FixtureMonkeyOptionsTest.java
Chanwon-Seo: Thanks for suggesting the test case. I'll add it to the code.

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/ExpressionPathValidator.java
Chanwon-Seo: Thank you for pointing out the missing part. I will fix it.

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/ExpressionPathValidator.java
```diff
@@ -0,0 +1,56 @@
...
+
+	public static boolean isValidFieldPath(
+		Class<?> rootClass, List<NextNodePredicate> predicates, PropertyNameResolver propertyNameResolver
+	) {
+		if (predicates == null || predicates.isEmpty()) {
+			return false;
+		}
+		Class<?> currentClass = Types.getActualType(rootClass);
+		for (NextNodePredicate predicate : predicates) {
+			if (predicate instanceof PropertyNameNodePredicate) {
```
seongahjo: It is better to continue early.

```java
if (!(predicate instance PropertyNameNodePredicate)){
    continue;
}
```

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/StrictModeNextNodePredicateContainer.java
```diff
@@ -0,0 +1,48 @@
...
+
+package com.navercorp.fixturemonkey.expression;
+
+import java.util.AbstractList;
+import java.util.List;
+
+import com.navercorp.fixturemonkey.api.property.PropertyNameResolver;
+import com.navercorp.fixturemonkey.tree.NextNodePredicate;
+
+public class StrictModeNextNodePredicateContainer extends AbstractList<NextNodePredicate> {
```
seongahjo: Since inheritance is not required, please add `final`.

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/StrictModeMonkeyExpressionFactory.java
```diff
@@ -0,0 +1,77 @@
...
+package com.navercorp.fixturemonkey.expression;
+
+import java.util.List;
+
+import com.navercorp.fixturemonkey.api.property.PropertyNameResolver;
+import com.navercorp.fixturemonkey.tree.ApplyStrictModeResolver;
+import com.navercorp.fixturemonkey.tree.NextNodePredicate;
+import com.navercorp.fixturemonkey.tree.NodeResolver;
+
+public class StrictModeMonkeyExpressionFactory implements MonkeyExpressionFactory {
```
seongahjo: Since inheritance is not required, please add 'final'.

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/FixtureMonkeyBuilder.java
```diff
@@ -498,21 +495,10 @@ public FixtureMonkeyBuilder enableLoggingFail(boolean enableLoggingFail) {
...
-
-				@Override
-				public List<NextNodePredicate> toNextNodePredicate() { // TODO:
-					return monkeyExpression.toNextNodePredicate();
-				}
-			};
-		};
+		this.monkeyExpressionFactory = new StrictModeMonkeyExpressionFactory(
+			new ArbitraryExpressionFactory(),
+			fixtureMonkeyOptionsBuilder.getDefaultPropertyNameResolver()
```
seongahjo: It could cause an issue if we use options such as the `pushAssignableTypePropertyNameResolver`.

I think it is better to add `propertyNameResolvers` in FixtureMonkeyBuilder as well. 

```
private List<MatcherOperator<PropertyNameResolver>> propertyNameResolvers = new ArrayList<>();
```

All options, such as `pushAssignableTypePropertyNameResolver`, use propertyNameResolvers from FixtureMonkeyBuilder and pass them to FixtureMonkeyOptionsBuilder within the build.

fixture-monkey/src/test/java/com/navercorp/fixturemonkey/test/FixtureMonkeyOptionsTest.java
Chanwon-Seo: ##### Original test case
> ```java
> .isThrownBy(() -> sut.giveMeBuilder(String.class)
> 				.size("nonExistentField", 1)
> ```

I've modified the test code because the `original test case` use of `String.class` did not meet the intended purpose.

Thank you for your review.

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/FixtureMonkeyBuilder.java
Chanwon-Seo: I have made improvements based on your review. 👍

In addition to the `pushAssignableTypePropertyNameResolver` you mentioned, I found that the same issue could affect the `defaultPropertyNameResolver`.

To address this, I've updated the code so that all `PropertyNameResolver` options are now applied consistently during the build process.

This guarantees uniform behavior regardless of the order in which the builder's options are called. I have also included additional test cases for this change.

Your review would be greatly appreciated.

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/StrictModeNextNodePredicateContainer.java
```diff
@@ -0,0 +1,48 @@
...
+import com.navercorp.fixturemonkey.tree.NextNodePredicate;
+
+public final class StrictModeNextNodePredicateContainer extends AbstractList<NextNodePredicate> {
+	private final List<NextNodePredicate> delegate;
+
+	public StrictModeNextNodePredicateContainer(
+		List<NextNodePredicate> delegate, Class<?> rootClass, PropertyNameResolver propertyNameResolver
+	) {
+		this.delegate = delegate;
+		if (!ExpressionPathValidator.isValidFieldPath(rootClass, delegate, propertyNameResolver)) {
```
seongahjo: To prevent the use of other classes, it is much better to inline `ExpressionPathValidator.isValidFieldPath` in `StrictModeNextNodePredicateContainer`.

I'm not sure it could be reused; it's better to extract it when needed.

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/FixtureMonkeyBuilder.java
```diff
@@ -587,4 +575,37 @@ public FixtureMonkey build() {
...
+
+	private void applyDeferredOptions() {
+		if (defaultPropertyNameResolver != null) {
+			fixtureMonkeyOptionsBuilder.defaultPropertyNameResolver(defaultPropertyNameResolver);
+		}
+
+		for (MatcherOperator<PropertyNameResolver> propertyNameResolver : propertyNameResolvers) {
+			fixtureMonkeyOptionsBuilder.insertFirstPropertyNameResolver(propertyNameResolver);
+		}
+	}
```
seongahjo: Inlining this method is preferable too.

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/FixtureMonkeyBuilder.java
```diff
@@ -587,4 +575,37 @@ public FixtureMonkey build() {
...
+
+	private PropertyNameResolver buildCompositePropertyNameResolver(
+		FixtureMonkeyOptions fixtureMonkeyOptions
+	) {
+		return property -> fixtureMonkeyOptions.getPropertyNameResolvers().stream()
+			.filter(it -> it.getMatcher().match(property))
+			.findFirst()
+			.map(it -> it.getOperator().resolve(property))
+			.orElseGet(() -> fixtureMonkeyOptions.getDefaultPropertyNameResolver().resolve(property));
+	}
```
seongahjo: To keep the consistency, it would be better to use the prefix of `new` instead of `build`.

This could cause confusion in the context of the FixtureMonkeyBuilder.

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/StrictModeNextNodePredicateContainer.java
Chanwon-Seo: Hi @seongahjo 
I appreciate your thorough review. I have made changes to the code based on your feedback on the shortcomings.

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/StrictModeNextNodePredicateContainer.java
```diff
@@ -0,0 +1,78 @@
...
+	public NextNodePredicate get(int index) {
+		return delegate.get(index);
+	}
+
+	@Override
+	public int size() {
+		return delegate.size();
+	}
+
+	private boolean isValidFieldPath(
```
seongahjo, line 69: I think there is a limitation in that it cannot check the element.

For example, it would not throw an exception in the following case:

```
String actual = sut.giveMeBuilder(new TypeReference<List<List<SimpleObject>>>() {
			})
			.size("$", 1)
			.size("$[0]", 1)
			.set("$[0][1].str", "expected")
			.sample()
			.get(0)
			.get(0)
			.getStr();
```

Nevertheless, it is a useful feature, so adding a comment to describe the limitation would be better.

Chanwon-Seo: > For example, it would not throw an exception in the following case:

here is a passing test case.

``` java
thenThrownBy(
		() -> sut.giveMeBuilder(new TypeReference<List<List<SimpleObject>>>() {
			})
			.size("$", 1)
			.size("$[0]", 1)
			.set("$[0][1].str", "expected")
			.sample()
	).isInstanceOf(IllegalArgumentException.class)
		.hasMessageContaining("No matching results for given NodeResolvers.");

```

As for the other test case you suggested, I felt it was already covered by the existing tests, so I didn't add it separately.
However, if you still think a more explicit test would be valuable, I'm happy to add it.

On a related note, I'm not sure about the best way to proceed with the isValidFieldPath method.
Should Iadd a comment to explain its current limitations, or would adding a test case that demonstrates the issue be sufficient?

If you think a comment is the right approach, could you please recommend what it should say?

seongahjo: Sorry, I've misunderstood. The above case was worked on using the existing `MonkeyExpressionFactory`.

I mean, in the case below, it does not throw an error even if the first element does not exist.

```
FixtureMonkey sut = FixtureMonkey.builder().useExpressionStrictMode().build();

		List<SimpleObject> actual = sut.giveMeBuilder(new TypeReference<List<List<SimpleObject>>>() {
			})
			.size("$", 1)
			.size("$[1]", 1)
			.sample()
			.get(0);
```

I think it's an edge case. It would be better to write a comment about the limitations.

Chanwon-Seo: Thanks for the late-night review!

I've addressed it in [commit](https://github.com/naver/fixture-monkey/pull/1223/commits/dee037053a71db61413662e7b9dd50e6bd41f4c7)

seongahjo (approved):
👍

PR #1235, opened 2025-09-04
Add JSpecify annotations in default annotation list

### LINKED ISSUE #1233: Support for JSpecify annotations in DefaultNullInjectGenerator

I'm trying to use JSpecify annotations to specify nullability in a new Java project. 
However, the current default annotation list used by `DefaultNullInjectGenerator` does not include JSpecify annotations.
So, I suggest `DefaultNullInjectGenerator` should recognize JSpecify's `@Nullable` and `@NonNull` annotations

## (Optional): Suggest A Solution

To recognize JSpecify annotations, we would need to retrieve annotations from `AnnotatedType`. 
When `DefaultNullInjectGenerator` fetches property annotations, it should check all annotations from `property.getAnnotatedType().getAnnotations()`.

```java
Set<String> annotationsFromType = Arrays.stream(context.getProperty().getAnnotatedType().getAnnotations())
          .map(it -> it.annotationType().getName())
          .collect(toSet());
```

JSpecify provides an additional `@NullMarked` annotation that can be applied to classes, packages, or modules. 
However, I'm not sure if this annotation can be recognized in the `generate` method of `DefaultNullInjectGenerator`.

[`@NullMarked` docs](https://jspecify.dev/docs/user-guide/#nullmarked)

Nevertheless, I suggest adding support for JSpecify's `@Nullable` and `@NonNull` annotations that are directly applied to properties.

*Yes*

jbl428:
## Summary

resolves #1233

## How Has This Been Tested?

`@NonNull` test

```java

	@Property
	void defaultNotNullNotWorksWhenSetDefaultNullInjectGenerator() {
			.defaultNullInjectGenerator(
				new DefaultNullInjectGenerator(
					ALWAYS_NULL_INJECT,
					false,
					false,
					false,
					new HashSet<>(DEFAULT_NULLABLE_ANNOTATION_TYPES),
					new HashSet<>(DEFAULT_NOTNULL_ANNOTATION_TYPES)
				)
			)
			.build();

		String actual = sut.giveMeOne(SimpleObject.class)
			.getStr();

		then(actual).isNotNull();
	}

	@Data
	public static class SimpleObject {
		private @org.jspecify.annotations.NonNull String str;
        }
```

`@Nullable` test

```java
	@Property
	void defaultNotNullNotWorksWhenSetDefaultNullInjectGenerator() {
			.defaultNullInjectGenerator(
				new DefaultNullInjectGenerator(
					ALWAYS_NULL_INJECT,
					false,
					true,
					false,
					new HashSet<>(DEFAULT_NULLABLE_ANNOTATION_TYPES),
					new HashSet<>(DEFAULT_NOTNULL_ANNOTATION_TYPES)
				)
			)
			.build();

		String actual = sut.giveMeOne(SimpleObject.class)
			.getStr();

		then(actual).isNull();
	}


	@Data
	public static class SimpleObject {
		private @org.jspecify.annotations.Nullable String str;
        }
```

### DIFF
--- a/buildSrc/src/main/kotlin/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/Dependencies.kt
@@ -19,6 +19,8 @@
 object Versions {
     const val FIND_BUGS_JSR305 = "3.0.2"
     const val FIND_BUGS_ANNOTATIONS = "3.0.1"
+    const val JSPECIFY = "1.0.0"
+    const val CHECKER_QUAL = "3.50.0"
     const val SLF4J = "1.7.25"
     const val LOGBACK = "1.2.9"
     const val JUNIT_ENGINE = "1.9.1"
--- a/fixture-monkey-tests/java-tests/build.gradle.kts
+++ b/fixture-monkey-tests/java-tests/build.gradle.kts
@@ -5,7 +5,9 @@ plugins {
 dependencies {
     testImplementation(project(":fixture-monkey-javax-validation"))
     testImplementation(project(":fixture-monkey-datafaker"))
-    testImplementation("org.projectlombok:lombok:${Versions.LOMBOK}")
     testImplementation(project(":fixture-monkey-jackson"))
+    testImplementation("org.jspecify:jspecify:${Versions.JSPECIFY}")
+    testImplementation("org.checkerframework:checker-qual:${Versions.CHECKER_QUAL}")
+    testImplementation("org.projectlombok:lombok:${Versions.LOMBOK}")
     testAnnotationProcessor("org.projectlombok:lombok:${Versions.LOMBOK}")
 }
--- a/fixture-monkey-tests/java-tests/src/test/java/com/navercorp/fixturemonkey/tests/java/DefaultNullInjectGeneratorTest.java
+++ b/fixture-monkey-tests/java-tests/src/test/java/com/navercorp/fixturemonkey/tests/java/DefaultNullInjectGeneratorTest.java
@@ -1,0 +1,83 @@
+/*
+ * Fixture Monkey
+ *
+ * Copyright (c) 2021-present NAVER Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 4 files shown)


fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/generator/DefaultNullInjectGenerator.java
```diff
@@ -43,7 +43,8 @@ public final class DefaultNullInjectGenerator implements NullInjectGenerator {
 			"org.springframework.lang.Nullable",
 			"org.checkerframework.checker.nullness.qual.Nullable",
 			"org.eclipse.jgit.annotations.Nullable",
-			"org.jmlspecs.annotation.Nullable"
+			"org.jmlspecs.annotation.Nullable",
+			"org.jspecify.annotations.Nullable"
```
jbl428, line 47: In version 1.1.14, we had to call `context.getProperty().getAnnotatedType().getAnnotations()` to recognize JSpecify annotations.
but starting from version 1.1.15, it seems like JSpecify annotations can now be recognized with just `context.getProperty().getAnnotations()`.

jbl428: I think this change might be caused by [this PR](https://github.com/naver/fixture-monkey/pull/1226) 
Am i right?

seongahjo: The values of the `context.getProperty().getAnnotatedType().getAnnotations()` and the `context.getProperty().getAnnotations()` should be the same in version 1.1.15. 

Are there any exceptions to this?

seongahjo: It might be helpful to add a test to the `fixture-monkey-tests` module.

You can add the dependency freely.

jbl428: > The values of the context.getProperty().getAnnotatedType().getAnnotations() and the context.getProperty().getAnnotations() should be the same in version 1.1.15.
> Are there any exceptions to this?

This was a confirmation question to check whether the issue could be resolved with just the current modifications 😄

jbl428: @seongahjo I've added the test code.

PR #1236, opened 2025-09-04
Fix strict mode validation for nested container expressions and generic types

### LINKED ISSUE #1234: "No matching results for given container expression" error in 1.1.15
## Describe the bug

When using the expression with `size(... into ..., ...)`, the error message appears:
`No matching results for given container expression.`

This issue does not occur in version **1.1.14** of Fixture Monkey, but it does occur in **1.1.15**.

## Your environment

* version of Fixture Monkey: 1.1.15 (works fine in 1.1.14)
* version of Java: 21
* version of Kotlin: 2.1.21

## Steps to reproduce

```kotlin
FIXTURE.giveMeKotlinBuilder<SampleParentDto>()
    .size(SampleParentDto::sections, 2)
    .size(SampleParentDto::sections[0] into SampleParentDto.Section::items, 3)
    .size(SampleParentDto::sections[1] into SampleParentDto.Section::items, 1)
    .sample()
```

Running the above code with Fixture Monkey **1.1.15** throws the error.

## Expected behaviour

The builder should correctly apply the `size(... into ..., ...)` constraints and generate an object without errors, just as it does in version **1.1.14**.

## Actual behaviour

In version **1.1.15**, the code fails with:
`No matching results for given container expression.`

Chanwon-Seo:
## Summary
Fixes strict mode validation for nested container expressions and generic types by ensuring correct type propagation.

Fixes #1234 

## (Optional): Description
Ensured `AnnotatedType` is correctly propagated for container types by using `ContainerElementPredicate` to traverse into their generic types. Refactored validation logic for clarity.

## How Has This Been Tested?
Verified with `FixtureMonkeyOptionsTest.strictModeNestedListValidExpression()`.

## Is the Document updated?
N/A

### DIFF
--- a/fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/StrictModeNextNodePredicateContainer.java
+++ b/fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/StrictModeNextNodePredicateContainer.java
@@ -23,6 +23,7 @@ import java.lang.reflect.Field;
 import java.util.AbstractList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 
 import com.navercorp.fixturemonkey.api.property.FieldProperty;
@@ -30,6 +31,10 @@ import com.navercorp.fixturemonkey.api.property.PropertyNameResolver;
 import com.navercorp.fixturemonkey.api.type.Types;
 import com.navercorp.fixturemonkey.tree.ContainerElementPredicate;
 import com.navercorp.fixturemonkey.tree.NextNodePredicate;
+import com.navercorp.fixturemonkey.tree.NodeAllElementPredicate;
+import com.navercorp.fixturemonkey.tree.NodeElementPredicate;
+import com.navercorp.fixturemonkey.tree.NodeKeyPredicate;
+import com.navercorp.fixturemonkey.tree.NodeValuePredicate;
 import com.navercorp.fixturemonkey.tree.PropertyNameNodePredicate;
 import com.navercorp.fixturemonkey.tree.StartNodePredicate;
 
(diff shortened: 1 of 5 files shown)


fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/StrictModeNextNodePredicateContainer.java
```diff
@@ -61,31 +64,46 @@ public int size() {
...
+						return resolvedFieldName.equals(fieldName);
+					})
+					.findFirst();
 
-			if (!field.isPresent()) {
+				if (!field.isPresent()) {
+					return false;
+				}
+
+				rootAnnotatedType = field.get().getAnnotatedType();
```
seongahjo: It is better to use the local variable than the parameter `rootAnnotatedType` for the traversal.

fixture-monkey/src/main/java/com/navercorp/fixturemonkey/expression/StrictModeNextNodePredicateContainer.java
```diff
@@ -61,31 +64,46 @@ public int size() {
...
+			} else if (predicate instanceof ContainerElementPredicate) {
```
seongahjo, line 103: It would be better to add the handling logic for the following predicates: `NodeValuePredicate`, `NodeKeyPredicate` and `NodeAllElementPredicate`.

The tests for the expressions `*` and `InnerSpec` refer to `Map` respectively.

Chanwon-Seo: I'm working on the StrictMode validation for InnerSpec and running into an issue with generic type resolution for Map values.

When validating a path into a generic Map value (e.g., Map<String, Item>), the Item type is being lost and treated as Object.

Do you have any initial ideas or a suggested approach for this?

seongahjo: @Chanwon-Seo 
Sorry for the late response.
Since the generic type is erased at runtime, it cannot be resolved using direct field reflection.

You don't need to check that the exact type of the Map is matched; you can match it by the key or value of the Map.
I think you can figure it out from `MapKeyElementProperty`, `MapValueElementProperty`

PR #1238, opened 2025-09-07
Add LongCombinableArbitrary

kail-is:
## Summary
Add LongCombinableArbitrary for customizable Long value generation following existing CombinableArbitrary pattern with Long-specific enhancements.

## Description
Implemented LongCombinableArbitrary to provide easy customization of Long value generation, addressing the lack of long-specific constraints in current API.

**Added components:**
- `LongCombinableArbitrary` interface with standard constraint methods
- `LongCombinableArbitraryDelegator` for filter/null injection operations  
- `JqwikLongCombinableArbitrary` implementation
- Factory method `CombinableArbitrary.longs()`

**Key features:**
- Standard numeric constraints: `.positive()`, `.negative()`, `.even()`, `.odd()`, `.withRange(long, long)`
- Long-specific `.nonZero()` method for avoiding division by zero errors
- Long-specific `.multipleOf(long)` method for generating multiples of specific values
- Last-method-wins behavior for constraint precedence (nonZero/multipleOf override previous constraints)
- Full integration with existing filter/map/unique operations

**Usage example:**
```java
Long nonZeroLong = CombinableArbitrary.longs().nonZero().combined();
Long multipleOfSeven = CombinableArbitrary.longs().multipleOf(7L).combined();
Long positiveLong = CombinableArbitrary.longs().withRange(1L, 1000L).positive().combined();
```

## How Has This Been Tested?
Added comprehensive test suite in `LongCombinableArbitraryTest`:
- Basic constraint validation (positive, negative, even, odd, nonZero, multipleOf)
- Range constraint testing with various boundaries
- Method chaining and last-method-wins precedence verification
- Integration with filter, map, injectNull, unique operations
- Complex constraint combinations and mathematical operation safety

## Is the Document updated?
Documentation will be updated to include LongCombinableArbitrary usage examples alongside existing numeric arbitrary types.

### DIFF
--- a/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/jqwik/JqwikLongCombinableArbitrary.java
+++ b/fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/jqwik/JqwikLongCombinableArbitrary.java
@@ -88,9 +88,6 @@ public final class JqwikLongCombinableArbitrary implements LongCombinableArbitra
 
 	@Override
 	public LongCombinableArbitrary multipleOf(long divisor) {
-		if (divisor == 0L) {
-			throw new IllegalArgumentException("divisor must not be zero.");
-		}
 		return new JqwikLongCombinableArbitrary(
 			Arbitraries.longs().map(it -> toMultipleOf(it, divisor))
 		);


fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/jqwik/JqwikLongCombinableArbitrary.java
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	@Override
+	public LongCombinableArbitrary negative() {
+		return new JqwikLongCombinableArbitrary(Arbitraries.longs().lessOrEqual(-1L));
+	}
+
+	@Override
+	public LongCombinableArbitrary even() {
+		return new JqwikLongCombinableArbitrary(Arbitraries.longs().filter(it -> it % 2 == 0));
```
seongahjo: @kail-is 
`filter` has a performance issue if it is not used with the Jqwik engine.

It is better to use the map to map odd and even values.
The range of longs should also be considered.

seongahjo: If you cannot use `map`, it is better to use `filter`.

fixture-monkey-kotest/src/main/kotlin/com/navercorp/fixturemonkey/kotest/KotestLongCombinableArbitrary.kt
```diff
@@ -0,0 +1,64 @@
...
+        KotestLongCombinableArbitrary(Arb.long().filter { it % 2 == 0.toLong() })
+
+    override fun odd(): LongCombinableArbitrary =
+        KotestLongCombinableArbitrary(Arb.long().filter { it % 2 != 0.toLong() })
+
+    override fun nonZero(): LongCombinableArbitrary =
+        KotestLongCombinableArbitrary(Arb.long().filter { it != 0.toLong() })
+
+    override fun multipleOf(divisor: Long): LongCombinableArbitrary =
+        KotestLongCombinableArbitrary(Arb.long().filter { it % divisor == 0.toLong() })
```
seongahjo: It is same as the `JqwikLongCombinableArbitrary`

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/jqwik/JqwikLongCombinableArbitrary.java
kail-is: Thanks for pointing that out. I’ve switched even/odd/nonZero/multipleOf to use map instead of filter, so the existing range constraints are preserved and we avoid the extra retry overhead.

fixture-monkey-kotest/src/main/kotlin/com/navercorp/fixturemonkey/kotest/KotestLongCombinableArbitrary.kt
kail-is: I’ve applied the same fix here as well, just like I did in `JqwikLongCombinableArbitrary`.

PR #1255, opened 2025-11-10
Add deterministic random class

210-reverof:
## Summary
  Implement DeterministicRandom class with seed-based Random instance caching for deterministic test reproducibility.

## Description
  - Created `DeterministicRandom` class that extends `Random`
    - Automatically caches Random instances using `nextLong()` generated values as keys
    - Provides `getRandomInstance(long seed)` to retrieve cached Random instances by seed
    - Provides `getCurrentSeedRandom()` to get the Random instance for the current seed
    - Implements cache size management with a limit of 32 entries to prevent memory leaks
  - Added 7 comprehensive property-based tests using JQwik to verify caching behavior and requirements

## How Has This Been Tested?
  - Each test runs 1000 iterations with different seed values (7000 total test cases)
  - Build verification: `./gradlew :fixture-monkey-api:test` passes successfully
  - Verified: cache creation on `nextLong()`, instance retrieval, cache size limit (≤32), deterministic behavior, and seed independence

## Is the Document updated?
  Not yet.


fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/DeterministicRandom.java
```diff
@@ -0,0 +1,70 @@
...
+package com.navercorp.fixturemonkey.api.random;
+
+import java.util.Map;
+import java.util.Random;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.apiguardian.api.API;
+import org.apiguardian.api.API.Status;
+
+@API(since = "1.1.16", status = Status.MAINTAINED)
```
seongahjo: It should be `EXPERIMENTAL`

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/DeterministicRandom.java
```diff
@@ -0,0 +1,70 @@
...
+		this.currentSeed = seed;
+	}
+
+	@Override
+	public long nextLong() {
+		long value = super.nextLong();
+		this.currentSeed = value;
+
+		if (!SEED_CACHE.containsKey(value)) {
+			manageCacheSize();
```
seongahjo: I don't think it's necessary to extract the method; it would be better to inline it.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/DeterministicRandom.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.Map;
+import java.util.Random;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.apiguardian.api.API;
+import org.apiguardian.api.API.Status;
+
+@API(since = "1.1.16", status = Status.MAINTAINED)
+public class DeterministicRandom extends Random {
```
seongahjo: I think it would be better to create a new interface for obtaining the `Random` instance and the seed value.

It is to avoid using the `DeterministicRandom` directly.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/DeterministicRandom.java
210-reverof: Thank you for checking! I [fixed it!](https://github.com/naver/fixture-monkey/pull/1255/commits/84e0308c387dee1bbf50d8cf6fcffed61ca5a528)

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/DeterministicRandom.java
210-reverof: Thank you for your feedback! I created a new class(DeterministicRandomSource), which resolves this issue as well.

fixture-monkey-api/src/main/java/com/navercorp/fixturemonkey/api/random/DeterministicRandom.java
210-reverof: Thank you for your feedback! 🙂

  I created a new interface named RandomSource to obtain Random instances and seed values.
  I also implemented DeterministicRandomSource that encapsulates the caching logic,
  so the internal implementation is no longer exposed. 

[84e0308](https://github.com/naver/fixture-monkey/pull/1255/commits/84e0308c387dee1bbf50d8cf6fcffed61ca5a528)

seongahjo: @210-reverof 
I think `DeterministicRandom` is also needed. 
It will be an implementation of `RandomSource`

PR #1256, opened 2025-11-11
Add compile-time null safety

### LINKED ISSUE #1251: Add compile-time null safety with @NullMarked and NullAway

1. **Adding package-level null safety configuration** - Add `package-info.java` files with `@NullMarked` annotation from JSpecify to establish non-null as the default for all elements in each package
2. **Integrating NullAway static analysis** - Add NullAway as a compile-time checker to automatically detect potential null pointer exceptions and enforce proper null handling throughout the codebase

**ref:**
- https://docs.spring.io/spring-framework/reference/7.0/core/null-safety.html
- https://github.com/spring-projects/spring-boot/issues/46587

## (Optional): Suggest A Solution

**Proposed implementation approach:**

- Add `package-info.java` files to all existing packages with `@NullMarked` annotation
- NullAway integration

**Example package-info.java:**
```java
@NullMarked
package com.example.project.core;

import org.jspecify.annotations.NullMarked;
```

**Yes**

jbl428:
## Summary

resolves #1251

Added CheckerFramework to enable compile-time null-safety checking and improve null safety validation.
Added `@Nullable` annotations across the codebase to explicitly express nullability and strengthen null-safety.

## (Optional): Description

- Added SuppressWarnings to skip null checks for existing code that wasn't handling null validation properly.
- Added `@Nullable` annotations where it seemed safe to do so - please review to make sure this won't cause any issues.
- Disabled nullability checks for test files.
-  Changed some Kotlin types from nullable to non-null - please double-check that these changes are safe.

- After applying checkFramework, the compiler now shows warnings when `@Nullable` annotations aren't placed directly next to the type:
I actually addressed this in a previous PR, but it seems the commit got lost during conflict resolution. I'll handle this in a separate PR.

```java
@Nullable
private Boolean optionValidOnly; // invalid

private @Nullable Boolean optionValidOnly; // valid
```

## How Has This Been Tested?

Run tests

## Is the Document updated?

No

### DIFF
--- a/fixture-monkey-tests/kotlin-tests/src/test/kotlin/com/navercorp/fixturemonkey/tests/kotlin/KotlinInnerSpecTest.kt
+++ b/fixture-monkey-tests/kotlin-tests/src/test/kotlin/com/navercorp/fixturemonkey/tests/kotlin/KotlinInnerSpecTest.kt
@@ -182,7 +182,7 @@ class KotlinInnerSpecTest {
                 .setKotlinInner {
                     property("strMap") {
                         minSize(1)
-                        key {}
+                        key(null)
                     }
                 }
                 .sample()
--- a/fixture-monkey-kotlin/src/main/kotlin/com/navercorp/fixturemonkey/kotlin/KotlinInnerSpec.kt
+++ b/fixture-monkey-kotlin/src/main/kotlin/com/navercorp/fixturemonkey/kotlin/KotlinInnerSpec.kt
@@ -98,7 +98,7 @@ class KotlinInnerSpec {
      *
      * @param key value of the map key to set
      */
-    fun key(key: Any): KotlinInnerSpec {
+    fun key(key: Any?): KotlinInnerSpec {
         return applyToSpec { it.key(key) }
     }
 
@@ -116,7 +116,7 @@ class KotlinInnerSpec {
      *
      * @param keys The keys to set in the map. Can be empty.
      */
-    fun keys(vararg keys: Any): KotlinInnerSpec {
+    fun keys(vararg keys: Any?): KotlinInnerSpec {
         return applyToSpec { it.keys(*keys) }
     }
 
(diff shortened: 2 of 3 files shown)


fixture-monkey-tests/kotlin-tests/src/test/kotlin/com/navercorp/fixturemonkey/tests/kotlin/KotlinInnerSpecTest.kt
```diff
@@ -175,21 +175,6 @@ class KotlinInnerSpecTest {
         then(actual.strMap.containsValue(null as String?)).isTrue()
     }
 
-    @Property
-    fun keyNullThrows() {
```
seongahjo, line 179: Should it be removed? It would be better if it remained.

jbl428: In the following code, `mapKey` is recognized as not null, which causes a compilation error in Kotlin, so I removed the null check.
```java
private void setMapKey(Object mapKey) {
    if (mapKey == null) {
       throw new IllegalArgumentException(
          "Map key cannot be null."
       );
    }

    setValue(
       this.declarativeExpression.element(entrySize - 1).key(),
       mapKey
    );
}
```

Would it be better to declare `mapKey` as nullable to handle this issue?

I have one question: since passing null to the key would result in an error, I thought it would be better to prevent null from being passed at compile time. Is there a specific reason why null is allowed in this case?

seongahjo: @jbl428 
> Would it be better to declare mapKey as nullable to handle this issue?

Yes, that would be nice.

> Is there a specific reason why null is allowed in this case?

As the `InnerSpec` is the public API, users can enter any value, including null. Therefore, all parameters in the `InnerSpec` should be marked as `@Nullable`.

jbl428: https://github.com/naver/fixture-monkey/pull/1256/commits/535b3f0db9f8ed405cace86436d3f3de881870a4

Done!

seongahjo (approved):

PR #1283, opened 2026-05-22
Optimize Kotlin primary constructor fixture generation

char-yb:
## Summary

This PR optimizes Kotlin primary constructor-based fixture generation without adding new dependencies or changing public APIs.

It reduces per-sample overhead in Kotlin object creation paths, which can become more visible when tests run under JVM instrumentation tools such as Kover.

## (Optional): Description

The previous implementation always used `callBy(...)` and built intermediate maps for every generated Kotlin object. `callBy(...)` is still necessary when optional constructor parameters should be omitted, but it is more expensive than needed when all parameters are already available.

This PR keeps the existing optional-parameter behavior intact and only uses the faster `KFunction.call(*args)` path when every constructor parameter can be provided.

Changes include:

- Reduce repeated `actualType()` / `cachedKotlin()` calls in `PrimaryConstructorArbitraryIntrospector.match()`.
- Avoid `mapKeys` allocation in the constructor invocation path by building the name-value map directly.
- Use `KFunction.call(*args)` when no optional constructor parameter needs to be skipped.
- Keep `callBy(...)` when optional parameters must be skipped.
- Preserve explicit `null` behavior for nullable optional parameters.
- Change constructor parameter name lookup from `List.contains` to `Set.contains`.
- Add regression tests for optional default values and nullable optional default values.

Local quick JMH benchmark:

| Version | Score |
| --- | ---: |
| Before | 1214.915 ms/op |
| After | 904.979 ms/op |

Benchmark command:

```bash
./gradlew :fixture-monkey-benchmarks:fixture-monkey-benchmark-kotlin:jmh \
  -Pjmh.quick \
  '-Pjmh.include=.*primaryConstructorGenerateKotlinOrderSheetWithFixtureMonkey.*'
```

## How Has This Been Tested?

```
./gradlew :fixture-monkey-kotlin:test :fixture-monkey-kotlin:ktlintCheck
./gradlew :fixture-monkey-tests:kotlin-tests:test
```

I also verified the benchmark above locally to compare the Kotlin primary constructor fixture generation path before and after this change.

## Is the Document updated?
No.

This PR does not change public APIs, user-facing behavior, configuration, or documented usage. It only optimizes the existing Kotlin primary constructor generation path.

Please feel free to share any feedback, especially if there are concerns about the constructor invocation behavior or benchmark scope.

### DIFF
--- a/fixture-monkey-kotlin/src/main/kotlin/com/navercorp/fixturemonkey/kotlin/introspector/PrimaryConstructorArbitraryIntrospector.kt
+++ b/fixture-monkey-kotlin/src/main/kotlin/com/navercorp/fixturemonkey/kotlin/introspector/PrimaryConstructorArbitraryIntrospector.kt
@@ -36,15 +36,23 @@ import org.apiguardian.api.API
 import org.apiguardian.api.API.Status.MAINTAINED
 import org.slf4j.LoggerFactory
 import java.lang.reflect.Modifier
+import kotlin.reflect.KFunction
 import kotlin.reflect.KParameter
 
 @API(since = "0.4.0", status = MAINTAINED)
 class PrimaryConstructorArbitraryIntrospector : ArbitraryIntrospector, Matcher {
-    override fun match(property: Property): Boolean =
-        property.type.actualType().isKotlinType() &&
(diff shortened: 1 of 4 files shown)


seongahjo (approved):
👍

